Issue #5, opened 2016-02-04
Room for performance improvement in `split_by`


ztdwu:
The performance of `split_by` can be improved by using a more efficient tokenizing algorithm as written here (not by me) http://tcbrindle.github.io/posts/a-quicker-study-on-tokenising. I've included an inlined version of the algorithm in the test program below. Here are the execution times of splitting a short string 5,000,000 times, compiled with `-O3` on `clang 3.7` on x86_64 Linux 4.3:

```
Time taken: 19.8554 (fplus::split_words)
Time taken: 2.02921 (improved version)
```

```c++
#include <algorithm>
#include <chrono>
#include <iostream>
#include <vector>
#include "fplus.h"

// Slightly modified code based on http://tcbrindle.github.io/posts/a-quicker-study-on-tokenising/
template <typename Str, typename Container = std::vector<std::string>>
Container tokenize(const Str &str, const char delim = ' ') {
    Container result;
    auto inserter = std::inserter(result, std::begin(result));

    auto token_start = std::begin(str);
    while (token_start != std::end(str) && *token_start) {
        const auto token_end = std::find(token_start, std::end(str), delim);
        if (token_start != token_end) {
            *inserter = { token_start, token_end };
        }
        if (token_end == std::end(str)) break;
        token_start = std::next(token_end);
    }
    return result;
}


int main() {
    using namespace std::chrono;

    auto words = std::string{ "a bb ccc dddd eeeee ffffff ggggggg hhhhhhhhh" };

    const auto start_time = high_resolution_clock::now();
    for (auto _ = 0; _ < 5000000; _++) {
        // auto tokens = fplus::split_words(words);
        auto tokens = tokenize(words);
    }
    const auto elapsed = high_resolution_clock::now() - start_time;
    std::cout << "Time taken: " << duration_cast<duration<double>>(elapsed).count() << std::endl;
}
```

Dobiasd:
Wow, that is very cool improvement. Thank you very much!

`split_words` internally uses the more general [`split_by`](https://github.com/Dobiasd/FunctionalPlus/blob/c2d00f66eb736fa04249d8c1afc74dcedd41d305/include/fplus/split.h#L106). Are you interested in reimplementing `split_by` using this algorithm and submitting a pull request? Otherwise please let me know, I will then see what I can do. :)

ztdwu:
Sounds good. A little busy lately so I'll probably submit the PR in a few days, if that works for you.

[...]

PR #6, opened 2016-02-05
Add function split_words_by

ztdwu:
This PR address Issue #5 

A few notes:

1. It turns out that the original `split_by` function was not *that* inefficient compared to the new version (it was slower but not 10x slower). What was actually the problem was that `split_words` calls `split_by` like this: `split_by(logical_not(is_letter_or_digit<String>), false, str);`
This causes the predicate to check for everything that's not a letter or digit, meaning it has to compare against many, many ascii characters <del>(or more if widechar is used)</del> for each character in `str`. This is an unnecessary overhead in cases where users only want to split by a single character.

2. A function `split_words_by` is added to accept a string and a delimiter, and splits by comparing against what's *not* equal to the delimiter.

3. `split_words_by` has an overload that takes a range of delimiters to split on. 

4. I also added a default parameter `bool allowEmpty = false` to `split_words` and `split_words_by` to give more flexibility to the users in letting them choose whether to keep empty tokens. It's defaulted to false to preserve backwards compatibility.

5. There was an issue with the comment describing `split_words`. It said `// split_words("How are you?") == ["How", "are", "you?"]`, which isn't true because your `split_words` will remove all non-digit and non-letter characters, so "you?" will become "you".

6. `string_tools.h` has Windows-style line endings while `test.cpp` and `split.h` have Unix-style line endings. I kept the endings consistent to the files, but that was weird. :P

Here are the new benchmarks (note that this is run on a more powerful machine compared to Issue #5, so the baseline is different)
```
Time taken: 16.0496s (fplus::split_words)
Time taken: 1.60925s (fplus::split_words_by)
```
```c++
#include <chrono>
#include <iostream>
#include "fplus.h"

template <typename Splitter>
void measure(const std::string &label, const Splitter &splitter) {
    using namespace std::chrono;

    const auto start_time = high_resolution_clock::now();
    for (auto _ = 0; _ < 5000000; _++) {
        splitter();
    }
    const auto elapsed = high_resolution_clock::now() - start_time;
    std::cout << "Time taken: " << duration_cast<duration<double>>(elapsed).count();
    std::cout << "s " << label << std::endl;
}

int main() {
    const auto words = std::string{ "a bb ccc dddd eeeee ffffff ggggggg hhhhhhhhh" };
    measure("(fplus::split_words)",    [words]() { fplus::split_words(words); });
    measure("(fplus::split_words_by)", [words]() { fplus::split_words_by(words, ' '); });
}
```

### DIFF
--- a/include/fplus/split.h
+++ b/include/fplus/split.h
@@ -106,33 +106,21 @@ template <typename UnaryPredicate, typename ContainerIn,
 ContainerOut split_by
         (UnaryPredicate pred, bool allowEmpty, const ContainerIn& xs)
 {
-    typedef typename ContainerIn::value_type T;
-    typedef typename ContainerOut::value_type ContainerOutInner;
     check_unary_predicate_for_container<UnaryPredicate, ContainerIn>();
     static_assert(std::is_same<ContainerIn, typename ContainerOut::value_type>::value, "Containers do not match.");
+
     ContainerOut result;
     auto itOut = get_back_inserter(result);
-    ContainerOutInner current;
-    auto itOutCurrent = get_back_inserter(current);
-    for (const T& x : xs)
+    auto start = std::begin(xs);
+    while (start != std::end(xs) && *start)
     {
-        if (pred(x))
+        const auto stop = std::find_if(start, std::end(xs), pred);
+        if (start != stop || allowEmpty)
         {
-            if (is_not_empty(current) || allowEmpty)
-            {
-                *itOut = current;
-                current.clear();
-                itOutCurrent = get_back_inserter(current);
-            }
+            *itOut = { start, stop };
         }
-        else
-        {
-            *itOutCurrent = x;
-        }
-    }
-    if (is_not_empty(current) || allowEmpty)
-    {
-        *itOut = current;
+        if (stop == std::end(xs)) break;
+        start = std::next(stop);
     }
     return result;
 }
--- a/include/fplus/string_tools.h
+++ b/include/fplus/string_tools.h
@@ -48,12 +48,39 @@ String clean_newlines(const String& str)
         replace_tokens(String("\r\n"), String("\n"), str));
 }
 
-// Splits a string by the found whitespace characters.
-// split_words("How are you?") == ["How", "are", "you?"]
+// Splits a string by non-letter and non-digit characters.
+// split_words("How are you?") == ["How", "are", "you"]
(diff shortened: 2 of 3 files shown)


Dobiasd:
Thank you a lot for this perfect contribution! It is very valuable.

I only changed very minor things.
* The second version of `split_words_by` was renamed to `split_words_by_many`. I try to avoid overloads to make it easier to instantiate the function template when used as higher-order functions for things like `transform` etc.
* `allowEmpty` no longer has a default. This breaks backward compatibility, but makes is consistent to the other functions taking an `allowEmpty`.

Additionally I fixed the line ending mess I created. ;)

ztdwu:
Glad to help :)

ztdwu:
Btw I just realized that I accidentally copy-pasted the function description of `split_words` into the new functions without changing them, so both `split_words_by` and `split_words_by_many` have a comment saying `// Splits a string by non-letter and non-digit characters`, when they really should say `// Splits a string by the given delimiter(s)`. This is probably not worth of another PR, so maybe you can fix that alongside the next commit. Oops.

Dobiasd:
Just fixed it. Thanks for caring so much.

Dobiasd:
There was a problem with the new version of `split_by`. It did not work for containers having an element evaluating to zero at the beginning or after a dividing element. I guess this happened because the algorithm was initially made for C strings. But since C++ Strings are allowed to contain the char `\0`, and of course since `split_by` also has to work not only for strings, I just changed it with [this commit](https://github.com/Dobiasd/FunctionalPlus/commit/477f2adc55a03107ba5e7823d5b01a5b94a3b949).

Also the new version yielded a different result for things like `split(2, true, IntVector{2})`. The old version returned `{{},{}}`. The new version only returned `{{}}`. Of course it is debatable which one really is the correct result, but since [`splitWhen` from Haskell's `Data.List.Split`](https://hackage.haskell.org/package/split-0.1.1/docs/Data-List-Split.html#v:splitWhen) works this way (`splitWhen (==2) [2] === [[],[]]`), I'd like to have `fplus::split_by` to be the same. This was changed with [this commit](https://github.com/Dobiasd/FunctionalPlus/commit/847b1ddac93d50bfeb9fb82ca0f6c16205a79916).

ztdwu:
Ah good catch, I should have thought of those cases!

Dobiasd:
No worries. I just wanted to inform you. I'm still very happy with the new algorithm though.

PR #8, opened 2016-06-06
Add cmake to build and run tests

pfultz2:
This adds an initial cmake. On unix based system it can build, run the tests, and install like this:

```
mkdir build
cd build
# Set cxx flags if they are not already set by the toolchain
CXXFLAGS='-std=c++11' cmake ..
# Build and run the tests
cmake --build . --target check
# Install the library
cmake --build . --target install
```

For MSVC, you would just omit the `CXXFLAGS='-std=c++11'` setting.

Also, with standard cmake packaging support this can be installed with [cget](https://github.com/pfultz2/cget/) as well:

```
# Setup up toolchain to use c++11
cget init --std=c++11
# Test and install
cget install --test Dobiasd/FunctionalPlus
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,25 @@
+
+cmake_minimum_required(VERSION 2.8)
+
+install (DIRECTORY include/fplus DESTINATION include)
+install (FILES include/fplus.h DESTINATION include)
+
+include(CTest)
+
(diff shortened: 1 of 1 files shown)


Dobiasd:
Very cool. Thanks a lot!

pfultz2:
> Would you like to update the install instructions in README.md too?

Sure, I'll try to send something for that too.

PR #9, opened 2016-06-06
Add install instructions

pfultz2:
This adds the install instructions using cmake or cget.


Dobiasd:
Great! Thanks again. :)

PR #12, opened 2016-06-09
CMake for C++11 support

offa:
Checks if the compiler supports either `-std=c++11` or `-std=0x`<sup>1)</sup> and adds the compilerflag if available. If none of them a warning is shown.

`CXXFLAGS='-std=c++11'` is not required any more, it's detected automatically now:

```
mkdir build
cmake ..
make check
make install
```

If the warning is a problem for VC++, it's possible to do the test only for gcc / clang.


Dobiasd:
Thanks again. :)

PR #15, opened 2016-06-10
Added comments to nub and unique

CrikeeIP:


Dobiasd:
Thanks a lot. Contributions like these are very valuable in terms of usability.

PR #21, opened 2016-06-18
Adding variadic fplus::min and fplus::max to numeric.h

### LINKED ISSUE #17: Adding variadic math functions like min, max etc.
I have seen your math functions like `min_2`, `min_3` etc. and I wondered why not use variadic templates to remove duplicated code?

Inspired by http://articles.emptycrate.com/2016/05/14/folds_in_cpp11_ish.html I have made some benchmarks and it came out that variadic versions are faster on `-g` and on pair with optimizations with your versions.

```cpp
#include <vector>
#include <functional>
#include <utility>
#include <type_traits>
#include <iostream>
#include <chrono>

template <typename TimeT = std::chrono::milliseconds>
struct measure
{
  template <typename F, typename... Args>
  static typename TimeT::rep execution(F func, Args&&... args)
  {
    auto start = std::chrono::system_clock::now();

    // Now call the function with all the parameters you need.
    func(std::forward<Args>(args)...);

    auto duration = std::chrono::duration_cast<TimeT>(std::chrono::system_clock::now() - start);

    return duration.count();
  }
};

template <typename U, typename... V>
auto min(const U& u, const V&... v)
{
  using rettype = typename std::common_type_t<U, V...>;
  rettype result = static_cast<rettype>(u);
  (void)std::initializer_list<bool>{((v < result) ? (result = static_cast<rettype>(v), false) : false)...};
  return result;
}

// API search type: min_2 : (a, a) -> a
// Minimum of two values.
template <typename X>
const X& min_2(const X& a, const X& b)
{
  return std::min(a, b);
}

// API search type: min_3 : (a, a, a) -> a
// Minimum of three values.
template <typename X>
const X& min_3(const X& a, const X& b, const X& c)
{
  return min_2(min_2(a, b), c);
}

// API search type: min_4 : (a, a, a, a) -> a
// Minimum of four values.
template <typename X>
const X& min_4(const X& a, const X& b, const X& c, const X& d)
{
  return min_2(min_3(a, b, c), d);
}

// API search type: min_5 : (a, a, a, a, a) -> a
// Minimum of five values.
template <typename X>
const X& min_5(const X& a, const X& b, const X& c, const X& d, const X& e)
{
  return min_3(min_3(a, b, c), d, e);
}

int main()
{
  constexpr int COUNT = 50000000;
  std::vector<int> v, v1;
  v.reserve(COUNT);
  v1.reserve(COUNT);

  std::cout << "variadic version: " << '\n';
  std::cout << measure<std::chrono::milliseconds>::execution([&v]()
                                                             {
                                                               for (int i = -COUNT; i < COUNT; ++i)
                                                                 v.emplace_back(min(1, -14, 3, -1, i));

                                                             }) << "ms\n";

  std::cout << "functional plus version: " << '\n';
  std::cout << measure<std::chrono::milliseconds>::execution([&v1]()
                                                             {
                                                               for (int i = -COUNT; i < COUNT; ++i)
                                                                 v1.emplace_back(min_5(1, -14, 3, -1, i));

                                                             }) << "ms\n";
}
```

It both depends on data and optimizations but I have received the following results (not the measure of all times but hey... :) ):

```
g++ -g -Wall -Wpedantic -Wextra --std=c++14 main.cpp && ./a.out                                                                                                       
variadic version: 
2240ms
functional plus version: 
4335ms

g++ -O1 -Wall -Wpedantic -Wextra --std=c++14 main.cpp && ./a.out
variadic version: 
333ms
functional plus version: 
307ms

g++ -O2 -Wall -Wpedantic -Wextra --std=c++14 main.cpp && ./a.out
variadic version: 
326ms
functional plus version: 
347ms

g++ -O3 -Wall -Wpedantic -Wextra --std=c++14 main.cpp && ./a.out
311ms
functional plus version: 
321ms
```

The  variadic version is only worse on `O1` no idea why...

If anyone is happy with this type of approach I will be more than happy to submit a pull request for some of the functions in numeric which can be easily handled with variadic templates/initializer_list.

pmalek:
This will resolve #17 by adding `fplus:min()` and `fplus::max()`

### DIFF
--- a/include/fplus/numeric.h
+++ b/include/fplus/numeric.h
@@ -204,6 +204,17 @@ const X& min_5(const X& a, const X& b, const X& c, const X& d, const X& e)
     return min_3(min_3(a, b, c), d, e);
 }
 
+// API search type: min : (a, ...) -> a
+// Minimum of x number of values
+template <typename U, typename... V>
+auto min(const U& u, const V&... v) -> typename std::common_type<U, V...>::type
+{
+  using rettype = typename std::common_type<U, V...>::type;
+  rettype result = static_cast<rettype>(u);
+  (void)std::initializer_list<int>{((v < result) ? (result = static_cast<rettype>(v), 0) : 0)...};
+  return result;
+}
+
 // API search type: max_2 : (a, a) -> a
 // Maximum of two values.
 template <typename X>
@@ -234,6 +245,17 @@ template <typename X>
 const X& max_5(const X& a, const X& b, const X& c, const X& d, const X& e)
 {
     return max_3(max_3(a, b, c), d, e);
+}
+
+// API search type: max : (a, ...) -> a
+// Maximum of x number of values.
+template <typename U, typename... V>
+auto max(const U& u, const V&... v) -> typename std::common_type<U, V...>::type
+{
+  using rettype = typename std::common_type<U, V...>::type;
+  rettype result = static_cast<rettype>(u);
+  (void)std::initializer_list<int>{((v > result) ? (result = static_cast<rettype>(v), 0) : 0)...};
+  return result;
 }
 
 // API search type: cyclic_value : Float -> (Float -> Float)
--- a/test/tests.cpp
+++ b/test/tests.cpp
@@ -1381,6 +1381,40 @@ void Test_example_CollatzSequence()
     std::cout << collatz_dict[17] << std::endl;
 }
 
+void Test_VariadicNumericTools()
+{
+    using namespace fplus;
+
+    assert(min(1,2,3,4,5) == 1);
+    assert(min(1.01,1.02,1.03,1.04,1.05) == 1.01);
+    assert(min(-54,2,3,54,5) == -54);
+    assert(min(-54.2,2.7,3,54,5) == -54.2);
+    assert(min(123,123,123,124) == 123);
+    assert(min(123) == 123);
+    assert(min(123,123) == 123);
+    assert(min(123,123,123) == 123);
+    assert(min(-1) == -1);
+    assert(min(-1,-2) == -2);
+    assert(min(-1,-2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0) == -2);
+    assert(min('a','b','c') == 'a');
+
+    assert(max(1,2,3,4,5) == 5);
+    assert(max(1.01,1.02,1.03,1.04,1.05) == 1.05);
+    assert(max(-54,2,3,54,5) == 54);
+    assert(max(-54.2,2.7,3,54.85,5) == 54.85);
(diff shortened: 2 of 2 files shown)


Dobiasd:
This is a very elegant solution and avoids code repetition in the library.
The performance (compile time and run time) looks good too. Also having it C++11 compatible is nice, because I would like to keep this property.
Even binding parameters is nearly as easy as before

    `auto three_or_more = bind_1st_of_2(max_2<int>, 3);`

becomes

    `auto three_or_more = bind_1st_of_2(max<int, int>, 3);`

The only downside I can think of right now is the fact that the [API search type](www.editgym.com/fplus-api-search) `min : (a, ...) -> a` is not supported (yet?), since it does not fit into the usual Haskell-like type signature scheme. But I think I will simply change it to the one `max_2` currently has (`(a, a) -> a`). I guess users are most likely to find this particular function by name and not by type anyway. :)

Dobiasd:
Do you see a chance we could also use this method to unite `max_2_on`, `max_3_on`, `min_2_on`, `min_3_on`?

pmalek:
It is possible but with a slight twist to the API:

It would be necessary to include the functor/lamda/function pointer to e.g. `max_on`'s parameter list so that the number of parameters that this functor takes can be deduced. So you would have to call it like so:

        fplus::max_on(mod2, 4, 3);

instead of:

        fplus::max_2_on(mod2)(4, 3);

There would be a flaw to this change, namely you couldn't store result of `fplus::max_on` (being a lambda) for later usage on parameters that you would get later since because this couldn't deduce the number of parameters.

I have submitted a PR #30 with this change.

I believe there is a method to keep it "the way it is" but I haven't figured this out yet. Maybe there's a template magician reading this who can submit a patch without changing the API (return a lambda that will accept variable number of parameters instead of immediately returning the result).

Dobiasd:
Using `max_2_on` in the curried way is a feature I like quite a lot and it is used in the code base of my company multiple times. So I would love to preserve it, if possible with C++11 compatibility.

Please correct me if I misunderstood what you wrote.

The function `F f` `max_2_on` takes, always only takes one parameter. Every value then is transformed by this `f`.

[This SO answer](http://stackoverflow.com/a/16527660/1866775) indicates that at least the creation of lambdas with variable number of parameters should be possible. Perhaps we can find a way to return it too?

If this approach fails, we could still use the new variadic non-curried version and additionally keep the few currentl functions.

pmalek:
The problem here is that you want to return a lambda with variable number of arguments from a function that doesn't know the amount nor the count of the arguments of that lambda. This will be known when this returned lambda will be called.
I believe it is possible to write it but as of now I am not aware how to do this :)

Dobiasd:
OK, I understand. So be it. Let's [keep the few curried versions additionally to your new ones](https://github.com/Dobiasd/FunctionalPlus/pull/30) for now. :)

PR #23, opened 2016-06-18
fix: include `cmath` before using `fmod`

xtofl:


Dobiasd:
Thank you very much. :+1:

PR #29, opened 2016-06-19
replace `TEST_F`s by `TEST` (fixtures are not needed here)

xtofl:
This will simplify writing test cases.

Also factored out the travis build steps into simple bash scripts.  This will allow us to run the build easier on our local machines.

(Note: please only accept pull request after it's travis build has succeeded)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: generic
 
+sudo: required
+
 cache:
     apt: true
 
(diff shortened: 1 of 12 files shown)


Dobiasd:
Wow, this is extremely helpful. I'm quite new to travis ci and gtest, so this not only levels up the user UX of FunctionalPlus, but also gives me the opportunity to learn a lot.
Thank you very much.

PR #30, opened 2016-06-19
Adding variadic method fplus::min_on and fplus::max_on

pmalek:
This PR will allow for calls to `fplus::min_on` and `fplus::max_on` like functions with variable number of parameters.
There are 4 templates added in this PR:
* 2 for `min_on`: 1 for POD types that copies them (parameters) and 1 for non POD types (to use `const T&` types)
* 2 for `max_on`: 1 for POD types that copies them (parameters) and 1 for non POD types (to use `const T&` types)

There are additional `and_` and `are_all_pod` templates in `numeric.h` - I would say they don't belong there but I do not know where is the best place for them; if need be please move them.

### DIFF
--- a/include/fplus/numeric.h
+++ b/include/fplus/numeric.h
@@ -135,6 +135,107 @@ X int_power(X base, X exp)
     return base * int_power(base, exp - 1);
 }
 
+template <typename... Conds>
+struct and_ : std::true_type
+{
+};
+
+template <typename Cond, typename... Conds>
+struct and_<Cond, Conds...> : std::conditional<Cond::value, and_<Conds...>, std::false_type>::type
+{
+};
+
+template <typename... T>
+using are_all_pod = and_<std::is_pod<T>...>;
+
+// API search type: min_on : (a -> b) -> ((a, a) -> a)
+// minimum of x values after transformation for POD types
+// min_on(mod2, 4, 3) == 4
+// min_on(mod7, 3, 5, 7, 3) == 7
+template <typename F,
+          typename FirstT,
+          typename... FIn,
+          typename = typename std::enable_if<fplus::are_all_pod<FirstT, FIn...>::type>>
+auto min_on(F f, FirstT first, FIn... v) -> typename std::common_type<FirstT, FIn...>::type
+{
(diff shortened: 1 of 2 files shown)


include/fplus/numeric.h
```diff
@@ -135,6 +135,107 @@ X int_power(X base, X exp)
     return base * int_power(base, exp - 1);
 }
 
+template <typename... Conds>
+struct and_ : std::true_type
```
xtofl: Does that mean that `and_<>` is true?  In other words, I would expect `and` not to be defined for no conditions at all.  That can be expressed by just providing a forward declaration for `and_`, and supplying a case for a singleton:

    template<typename... Conds> struct and_; // fwd declaration only -> compile error for no conditions
    template<typename Cond> struct and_: public Cond {};
    template<typename Cond, typename Conds...> // as defined

Dobiasd: @xtofl : I do not know. But @pmalek probably can help. Anyway, It's worth noting that this part was removed from the code anyway by now.

pmalek: @xtofl You're right. This can be done e.g. like so:

```
#include <type_traits>
#include <string>

template <typename... Conds>
struct and_;

template <typename Cond>
struct and_<Cond> : Cond
{
};

template <typename Cond, typename... Conds>
struct and_<Cond, Conds...> : std::conditional<Cond::value, and_<Conds...>, std::false_type>::type
{
};

template <typename... T>
using are_all_pod = and_<std::is_pod<T>...>;

int main()
{
  static_assert(are_all_pod<int, float>::value, "");
  //static_assert(are_all_pod<int, float, std::string>::value, ""); - this will fail compilation
  static_assert(and_<std::true_type>::value, "");
  static_assert(and_<std::true_type,std::true_type>::value, "");
  //static_assert(and_<std::true_type,std::false_type>::value, ""); - this will fail compilation
  //static_assert(and_<std::false_type>::value, "");
  //and_<> a2; - this will fail compilation
}
```

Maybe there's a more elegant way but I cannot think of it right now.

PR #31, opened 2016-06-20
Some cmake improvements

offa:
Some cmake improvements.

@Dobiasd : What is the current version of the project? I have set `0.1` as placeholder, please set to a version that's suitable to you.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,4 @@
-
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 2.8.11)
 
 
 option(UNITTEST "Build unit tests" OFF)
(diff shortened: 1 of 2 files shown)


Dobiasd:
`0.1` is fine for now. As long as it does not begin with `1` we are free do whatever we like anyway, I guess. ;-)

Perhaps we can think about official release versions numbers (incl. release notes) in the future when the API is more stable.

offa:
Maybe it makes sense to release some unstable releases like v0.1, v0.2 etc., even before the api is stable?

Dobiasd:
The thing is I have no idea about the size of the user base and if anybody is really developing something using FunctionalPlus except the company I work for. :D

But perhaps it is good practice anyway, even with an unknown target group size.

PR #32, opened 2016-06-20
Typo fixed.

offa:
Typo fixed.


Dobiasd:
I think you are doing exceptional work here. Of course I am not talking about this particular fix here in isolation. But the fact that you even review and correct these small things is quite remarkable. :+1:

offa:
Thanks ;-)

PR #36, opened 2016-06-21
Adding curried versions of min_on and max_on

pmalek:
I believe this is the missing piece that we have been talking about in #21 :)

### DIFF
--- a/include/fplus/numeric.h
+++ b/include/fplus/numeric.h
@@ -190,7 +190,32 @@ auto min_on(F f, const FirstT& first, const FIn&... v) ->
   return result;
 }
 
-// API search type: max_on : ((a -> b), a, a) -> a
+template <typename F>
+struct min_on_t
+{
+  min_on_t(F _f) : f(_f) {}
+
+  template <typename T, typename... Ts>
+  auto operator()(T&& x, Ts&&... xs) -> typename std::common_type<T, Ts...>::type
+  {
+    return min_on(std::forward<F>(f), std::forward<T>(x), std::forward<Ts>(xs)...);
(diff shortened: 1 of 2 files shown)


Dobiasd:
That is awesome! I was expecting come very complicated template magic, but your solution is beautifully simple.

Of course I still respect the engineering art that you put into the non-curried versions, but to reduce code complexity and to avoid having overloaded functions (for binding, using with `transform`, `zip` etc) I think we should remove them now. Do you agree?

pmalek:
Well it all depends on your usage but sure. If you don't see their use then we can remove them.

Dobiasd:
Oh, I just realized, we can not remove them completely of course, because the curried version uses the non curried version. But I [moved it into the `internal` namespace](https://github.com/Dobiasd/FunctionalPlus/commit/935d584f887da729df44763599c6edcedd482cac).

Could you explain the difference between the versions for POD types and the version for non-POD types to me? Is there more to it than passing PODs per value and non-PODs per ref to const?

pmalek:
As of the current version there is no difference besides passing by lvalue/const ref.

I was wondering (and I still am) whether it would be better to use rvalue references in function signature there but I will have to get to higher level of expertise on them to reason about it:)

Dobiasd:
OK, thanks. An `int` or something alike should be faster passed by value than by ref to const, but right now this differentiation is made nowhere else in the library. I would love to have consistency here. Adapting everything else to this differentiation would increase the overall complexity a lot and make everything less DRY. So I think we should drop the pass-by-value versions. Do you approve this plan?

PR #40, opened 2016-06-24
GTest headers added as system includes to prevent Weffc++ warnings if

offa:
This is a small fix that adds gtest includes as system includes. If Gtest is not installed to system include paths, the `Weffc++` Warning prevents the FunctionalPlus tests from building.


offa:
:laughing: I was wondering why the build failed here, but worked on my Fork … then I noticed you've already merged it – therefore the PR was not available anymore --> build failed.

Dobiasd:
Haha. To me the commit in this PR looked like something, that would not break the travis build anyway, so I merged it right away without waiting for the travis PR results. :)

Issue #24, opened 2016-06-18
Adding Unit Tests?


offa:
What about adding unit tests to the project?

I have written some example tests, see here: https://github.com/offa/FunctionalPlus/tree/ut_gtest (`test` directory)

The tests are off per default at the moment – so no dependencies are needed. Only Building / running the unit tests requires gtest. To do so:

```
cd build
cmake -DUNITTEST=ON ..
make unittest
```

You can check the CI log's for the branch to see how it works.

Btw. I stumbled over #20 by writing an UT for it :+1:

Dobiasd:
This seems like a good idea to me. The [current test coverage](https://github.com/Dobiasd/FunctionalPlus/blob/0cdf523244b5f8a4ba4f743bc81d061f12a31fba/test/tests.cpp) is far from perfect, as your [drop_if_with_idx issue](https://github.com/Dobiasd/FunctionalPlus/issues/20) showed.

Is this a task you would be interested in pursuing?

offa:
I can't add all tests, but at least I can start – and tests will evolve. PR follows.

What's already available in my branch:

- Tests are optional (default off), so no test's means no further dependencies
- Searching for Gtest
- Some Test Cases (for filter and generate)
- make target for running the tests
- CI will build / run the available tests

Dobiasd:
OK, great. Just merged your [commit](https://github.com/Dobiasd/FunctionalPlus/commit/afd2e17d976c7e242a4be9b85469aab93c995fc3).

Dobiasd:
Btw, I just [moved the numeric tests over to Google Test](https://github.com/Dobiasd/FunctionalPlus/commit/dd06a5916c3118e265d3daab0545eee2a0dcca06). Does the way I did it look OK to you?

offa:
Yes! :+1:

offa:
More tests for generate and filter come in a minute.

Dobiasd:
All old unit tests are now moved over to google test. Not everything is a thing of beauty, i.e. in many cases I simply used regular expressions to convert the `assert`s to `EXPECT_EQ`s, but at least it works correctly.

offa:
@Dobiasd  is the old `tests.cpp` needed anymore?

Dobiasd:
Not for the unit tests. It would be cool if you could remove if from our `CMakeLists.txt`.

offa:
[...]

PR #43, opened 2016-06-28
Examples moved and added as an optional make target (see #24).

offa:
Examples moved and added as an optional make target (see #24).

### What changed:

- `tests.cpp` renamed to `examples.cpp` and moved to project root dir
- `examples` added as the target to build `examples.cpp`; it's optional so `make`, `make all`, `make test` etc. wont build it
- The examples are not included in the test targets anymore
- `make test` / `ctest` will run unit tests only (using ctest)

### Build & run examples:

```
make examples
./examples
```
(called in `build/`)


offa:
Oh, I have formatted the compiler flags a bit to reduce line length _(cosmetic only)_.

Dobiasd:
[This change](https://github.com/Dobiasd/FunctionalPlus/commit/0cdf523244b5f8a4ba4f743bc81d061f12a31fba) makes the line

    install(FILES include/fplus.h DESTINATION include)

in `CMakeLists.txt` superfluous, right?

Dobiasd:
OK, I'm sure now it could be thrown out, so I [did it](https://github.com/Dobiasd/FunctionalPlus/commit/bca43de8e434f52b487d348379081011e30144cc).

PR #53, opened 2016-10-14
CMake: proper searching for doctest added

offa:
Some CMake related changes according doctest:

- Proper searching for doctest added 
- Include path set for test targets


Dobiasd:
Awesome!

PR #55, opened 2016-12-11
Adding a unordered_set_includes ant tests.

GustavooPaiva:
Since std::includes, std::set_intersection, std::set_difference, std::set_symmetric_difference requires a ordered container it won't work for the unordered_set containers. 
So i created a function to the set_intersection and also added some tests. I tried to keep the style of the project, tell me if I did anything wrong. If you like it I could try to make the others functions.

Edit: sorry for typo

### DIFF
--- a/include/fplus/sets.hpp
+++ b/include/fplus/sets.hpp
@@ -25,6 +25,20 @@ bool set_includes(const SetType& set1, const SetType& set2)
         std::begin(set2), std::end(set2));
 }
 
+// API search type: unordered_set_includes : (Unordered_Set a, Unordered_Set a) -> Bool
+// Checks if every element of the second unordered_set is also present in the first unordered_set.
+// Also known as is_subset_of.
+template <typename UnorderedSetType>
+bool unordered_set_includes(const UnorderedSetType& set1, const UnorderedSetType& set2)
+{
+    for(const auto& x : set2)
+    {
+        if (set1.find(x) == set1.end())
+            return false;
+    }
+    return true;
+}
+
 // API search type: set_merge : (Set a, Set a) -> Set a
 // Returns the union of two given sets.
 template <typename SetType>
--- a/test/container_common_test.cpp
+++ b/test/container_common_test.cpp
@@ -987,6 +987,22 @@ TEST_CASE("container_common_test, set")
     REQUIRE_EQ(set_symmetric_difference(intSet1, intSet2), intSet7);
     REQUIRE_EQ(set_intersection(intSet1, intSet2), intSet4);
     REQUIRE_EQ(sets_intersection(IntSets({intSet1, intSet2, intSet3})), intSet8);
+
+    typedef std::unordered_set<int> IntUnordSet;
+
+    IntUnordSet intUnordSet1 = {0,1,2,3};
+    IntUnordSet intUnordSet2 = {2,3,4,5};
+    IntUnordSet intUnordSet3 = {0,2};
+    IntUnordSet intUnordSet4 = {2,3};
+    IntUnordSet intUnordSet5 = {0,1};
+    IntUnordSet intUnordSet6 = {0,1,2,3,4,5};
+    IntUnordSet intUnordSet7 = {0,1,4,5};
+
+    REQUIRE_EQ(unordered_set_includes(intUnordSet1, intUnordSet3), true);
+    REQUIRE_EQ(unordered_set_includes(intUnordSet3, intUnordSet1), false);
+    REQUIRE_EQ(unordered_set_includes(intUnordSet1, intUnordSet2), false);
+    REQUIRE_EQ(set_merge(intUnordSet1, intUnordSet2), intUnordSet6);
+    REQUIRE_EQ(set_merge(intUnordSet1, intUnordSet3), intUnordSet1);
 }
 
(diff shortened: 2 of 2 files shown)


test/container_common_test.cpp
```diff
@@ -987,6 +987,22 @@ TEST_CASE("container_common_test, set")
     REQUIRE_EQ(set_symmetric_difference(intSet1, intSet2), intSet7);
     REQUIRE_EQ(set_intersection(intSet1, intSet2), intSet4);
     REQUIRE_EQ(sets_intersection(IntSets({intSet1, intSet2, intSet3})), intSet8);
+
+    typedef std::unordered_set<int> IntUnordSet;
```
xtofl: Shouldn't we do `using IntUnordSet = std::unordered_set<int>` these days?

test/container_common_test.cpp
```diff
@@ -987,6 +987,22 @@ TEST_CASE("container_common_test, set")
...
+
+    IntUnordSet intUnordSet1 = {0,1,2,3};
+    IntUnordSet intUnordSet2 = {2,3,4,5};
+    IntUnordSet intUnordSet3 = {0,2};
+    IntUnordSet intUnordSet4 = {2,3};
+    IntUnordSet intUnordSet5 = {0,1};
+    IntUnordSet intUnordSet6 = {0,1,2,3,4,5};
+    IntUnordSet intUnordSet7 = {0,1,4,5};
+
+    REQUIRE_EQ(unordered_set_includes(intUnordSet1, intUnordSet3), true);
```
xtofl: If you would use a literal construction here, the code reader doesn't have to skip his focus up and down to verify the test requirement.  On top of that, you don't have to think deeply 'is `intUnordSet7` even used?'.

Also, is it possible to test for trueness and falseness instead of equality?

    REQUIRE_FALSE( unordered_set_includes(Unordered{0, 2}, {0, 1, 2, 3}) );

include/fplus/sets.hpp
```diff
@@ -25,6 +25,20 @@ bool set_includes(const SetType& set1, const SetType& set2)
         std::begin(set2), std::end(set2));
 }
 
+// API search type: unordered_set_includes : (Unordered_Set a, Unordered_Set a) -> Bool
+// Checks if every element of the second unordered_set is also present in the first unordered_set.
+// Also known as is_subset_of.
+template <typename UnorderedSetType>
+bool unordered_set_includes(const UnorderedSetType& set1, const UnorderedSetType& set2)
+{
+    for(const auto& x : set2)
```
xtofl: This raw for loop is actually a `find_if`:

    auto first_not_included = std::find_if(set2.begin(), set2.end(),
      [&](auto &&x){ return set1.find(x) == set1.end(); });
    return first_not_included != set2.end();

xtofl:
Maybe there is a way to automatically deduce whether a container is ordered or not (at least for standard containers) with type traits.  This may result in a uniform API.  I'm not entirely sure how to achieve it, but it may look somewhat like this:

    template<typename Container> struct is_ordered {};

    template<typename K, typename V, typename P, typename A>
    template<> struct is_ordered< std::map<K, V, P, A> > { static constexpr bool value = true; };

    template<typename V, typename P, typename A>
    template<> struct is_ordered< std::set<V, P, A> > { static constexpr bool value = true; };

    template<typename K, typename V, typename P, typename A>
    template<> struct is_ordered< std::unordered_map<K, V, P, A> > { static constexpr bool value = false; };

    template<typename K, typename V, typename P, typename A>
    template<> struct is_ordered< std::unordered_set<K, V, P, A> > { static constexpr bool value = false; };

And then we can use SFINAE to select the proper function at compile time.



    template<typename C, bool sfinae = std::enable_if< is_ordered<C>::value >::value >
    bool set_includes(const C &lhs, const C &rhs) { ... } // the ordered variants

    template<typename C, bool sfinae = std::enable_if< !is_ordered<C>::value >::value >
    bool set_includes(const C &lhs, const C &rhs) { ... } // the unordered variants


Then the caller does not have to bother selecting a different function for the same functionality, allowing the `set_includes` function variants to be used in generic functions, too.  'cause As it is now, this is not the case.

test/container_common_test.cpp
GustavooPaiva: Indeed, I just followed the way tests were coded. I'll change that. :+1:

Dobiasd:
@GustavooPaiva Thanks a lot. If you want to, I would love to see you implement the other functions too.

@xtofl Thank you very much for the mentoring work here.

test/container_common_test.cpp
Dobiasd: Yes, `REQUIRE_FALSE` would be nicer.

Dobiasd:
@GustavooPaiva If you would like to create a second PR cleaning up the small things xtofl mentioned (raw loop -> `find_if` and `REQUIRE_FALSE(x)` instead of `REQUIRE_EQ(x, false)`) I'd be happy to merge it immediately.

@xtofl I Like the idea of having a uniform API for unordered/ordered set/map.

GustavooPaiva:
@Dobiasd Ok, when I get home, I'll make a second PR fixing those things.

@xtofl Thank you for the feedback, I'll try to improve those in the future.

PR #57, opened 2016-12-14
Adding missing unordered_set functions and refactoring the unordered_…

GustavooPaiva:
There we go, the missing functions. Also i tried to use tips from you guys, @xtofl and @Dobiasd.
Sadly it is not uniform as @xtofl proposed. I'll try to make it work in the next week or so.

### DIFF
--- a/include/fplus/sets.hpp
+++ b/include/fplus/sets.hpp
@@ -28,11 +28,13 @@ bool set_includes(const SetType& set1, const SetType& set2)
 // API search type: unordered_set_includes : (Unordered_Set a, Unordered_Set a) -> Bool
 // Checks if every element of the second unordered_set is also present in the first unordered_set.
 // Also known as is_subset_of.
-template <typename UnorderedSetType>
-bool unordered_set_includes(const UnorderedSetType& set1, const UnorderedSetType& set2)
+template <typename UnorderSetType>
+bool unordered_set_includes(const UnorderSetType& set1,
+                            const UnorderSetType& set2)
 {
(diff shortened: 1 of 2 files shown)


include/fplus/sets.hpp
```diff
@@ -31,12 +31,9 @@ bool set_includes(const SetType& set1, const SetType& set2)
...
 bool unordered_set_includes(const UnorderedSetType& set1, const UnorderedSetType& set2)
 {
-    for(const auto& x : set2)
-    {
-        if (set1.find(x) == set1.end())
-            return false;
-    }
-    return true;
+    auto first_not_included = std::find_if(set2.begin(), set2.end(),
+                                           [&](const typename UnorderedSetType::value_type& x) -> bool { return set1.find(x) == set1.end();});
```
xtofl: style: this line goes over the Github diff window.

Also, there's no need to the lambda argument type or return type (supposing we do C++14):

    [&](auto&& x) { return set1.find(x) == set1.end();};

Dobiasd: Up to now FunctionalPlus works with C++11 and does not need a C++14 compiler. I would like to keep it that way.

Dobiasd:
Thanks @GustavooPaiva. Could you try to keep the code lines in the library code at a maximum of 80 characters, please?
For unit test code and API search types this rule does not apply.

GustavooPaiva:
There you go, @Dobiasd.

Dobiasd:
:+1:

PR #77, opened 2017-06-25
Apt Cache setting removed from CI config

offa:
Apt Cache setting removed from CI config, as this is actually not supported by Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ language: generic
 dist: trusty
 sudo: required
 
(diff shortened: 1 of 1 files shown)


Dobiasd:
After 3 hours still `Hang tight, the log cannot be shown until the build has started.` for the GCC 7 test. I think I'll merge anyways. ;)

offa:
Strange, the new build of master doesn't have this issue.

PR #84, opened 2017-08-26
Feature/generic lambdas

theodelrieu:
Hello, I made `zip_with` work with generic lambdas/transparent function objects/etc ...,
after spending two days on making `invoke` fully work (I still have to tests some `volatile const&&` stuff, but was too lazy to do so before showing you the result first!

This PR is not meant to be merged at all, just to know if you have some ideas to improve the `static_assert` machinery introduced here, and to discuss some points/questions.

`invoke` was definitely fun to do :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,12 +35,10 @@ if( UNITTEST )
     subdirs(test)
 endif()
 
-
 set(CMAKE_CXX_STANDARD 14)
 
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
-
 
 add_executable(readme_examples EXCLUDE_FROM_ALL examples/readme_examples.cpp)
 target_link_libraries(readme_examples ${CMAKE_THREAD_LIBS_INIT})
--- a/include/fplus/detail/function_traits_asserts.hpp
+++ b/include/fplus/detail/function_traits_asserts.hpp
@@ -1,0 +1,55 @@
+// Copyright 2015, Tobias Hermann and the FunctionalPlus contributors.
+// https://github.com/Dobiasd/FunctionalPlus
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <fplus/detail/invoke.hpp>
+#include <fplus/function_traits.hpp>
+
+namespace fplus
+{
+namespace detail
+{
+template <typename T, typename...>
+struct function_traits_asserts;
+
+template <
+    typename,
+    typename F,
+    typename... Args,
+    typename std::enable_if<is_invocable<F, Args...>::value, int>::type = 0>
+constexpr bool trigger_static_asserts()
+{
+  return true;
+}
+
+template <typename Tag,
+          typename F,
+          typename... Args,
+          typename std::enable_if<has_function_traits<F>::value &&
+                                      !is_invocable<F, Args...>::value,
+                                  int>::type = 0>
+constexpr bool trigger_static_asserts()
+{
+  // don't perform checks if function_traits<F> doesn't exist
+  (void)function_traits_asserts<Tag, F, Args...>{};
+  return true;
+}
+
+template <typename,
+          typename F,
+          typename... Args,
+          typename std::enable_if<!has_function_traits<F>::value &&
+                                      !is_invocable<F, Args...>::value,
+                                  int>::type = 0>
+constexpr bool trigger_static_asserts()
+{
+  static_assert(sizeof(F) == 0,
+                "F is not a Callable, or its definition is ill-formed");
+  return true;
+}
+}
+}
--- a/include/fplus/detail/invoke.hpp
+++ b/include/fplus/detail/invoke.hpp
@@ -1,0 +1,275 @@
+// Copyright 2015, Tobias Hermann and the FunctionalPlus contributors.
+// https://github.com/Dobiasd/FunctionalPlus
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <functional>
+#include <type_traits>
+#include <utility>
+
+#include <fplus/detail/meta.hpp>
+
+// borrowed to libc++
+#define FPLUS_INVOKE_RETURN(...) \
+  ->decltype(__VA_ARGS__)        \
+  {                              \
+    return __VA_ARGS__;          \
+  }
+
+namespace fplus
+{
+namespace detail
+{
+// We need std::invoke to detect callable objects
+//
+// source:
+// http://en.cppreference.com/mwiki/index.php?title=cpp/utility/functional/invoke&oldid=82514
+template <typename U>
+static std::true_type is_refwrap_test(const std::reference_wrapper<U>&);
+
+template <typename U>
+static std::false_type is_refwrap_test(const U&);
+
+template <typename T>
+struct is_reference_wrapper : decltype(is_refwrap_test(std::declval<T>()))
+{
+};
+
+template <typename T, typename U = typename std::decay<T>::type>
+struct unwrap_reference_wrapper
+{
+  using type = T;
+};
+
+template <typename T, typename U>
+struct unwrap_reference_wrapper<T, std::reference_wrapper<U>>
+{
+  using type = U&;
+};
+
+template <typename T>
+using unwrap_reference_wrapper_t = typename unwrap_reference_wrapper<T>::type;
+
+// note: clang only triggers the second static_assert
+//      - static_assert(is_invocable<&base_class::non_const_method, const derived_class&>::value, "");
+//      - static_assert(is_invocable<&base_class::non_const_method, const base_class&>::value, "");
+// GCC triggers both. To workaround this clang bug, we have to manage cv correctness ourselves
+
+template <typename T>
+struct is_const_member_function : std::false_type
+{
+};
+
+// decay doesn't add pointer to abominable functions, don't bother writing them
+template <typename R, typename... Args>
+struct is_const_member_function<R(Args...) const> : std::true_type
+{
+};
+
+template <typename R, typename... Args>
+struct is_const_member_function<R(Args...) const&> : std::true_type
+{
+};
+
+template <typename R, typename... Args>
+struct is_const_member_function<R(Args...) const&&> : std::true_type
+{
+};
+
+template <typename R, typename... Args>
+struct is_const_member_function<R(Args...) const volatile> : std::true_type
+{
+};
+
+template <typename R, typename... Args>
+struct is_const_member_function<R(Args...) const volatile&> : std::true_type
+{
+};
+
+template <typename R, typename... Args>
+struct is_const_member_function<R(Args...) const volatile&&> : std::true_type
+{
+};
+
+template <typename T>
+struct is_volatile_member_function : std::false_type
+{
+};
+
+// decay doesn't add pointer to abominable functions, don't bother writing them
+template <typename R, typename... Args>
+struct is_volatile_member_function<R(Args...) volatile> : std::true_type
+{
+};
+
+template <typename R, typename... Args>
+struct is_volatile_member_function<R(Args...) volatile&> : std::true_type
+{
+};
+
+template <typename R, typename... Args>
+struct is_volatile_member_function<R(Args...) volatile&&> : std::true_type
+{
+};
+
+template <typename R, typename... Args>
+struct is_volatile_member_function<R(Args...) const volatile> : std::true_type
+{
+};
+
+template <typename R, typename... Args>
+struct is_volatile_member_function<R(Args...) const volatile&> : std::true_type
+{
+};
+
+template <typename R, typename... Args>
+struct is_volatile_member_function<R(Args...) const volatile&&> : std::true_type
+{
+};
+
+template <typename Object, typename Signature>
+struct has_correct_cv
+{
+  // if object has no cv, every method can be called
+  // else the method must have the same cv than the object
+  static constexpr bool value =
+      std::is_same<typename std::remove_cv<Object>::type, Object>::value ||
+      ((is_volatile_member_function<Signature>::value ==
+        std::is_volatile<Object>::value) &&
+       (is_const_member_function<Signature>::value ==
+        std::is_const<Object>::value));
+};
+
+// pointer to member function - reference to object
+template <
+    typename Base,
+    typename T,
+    typename Derived,
+    typename... Args,
+    typename Unwrapped = unwrap_reference_wrapper_t<Derived>,
+    typename std::enable_if<
+        is_function<T>::value &&
+            has_correct_cv<typename std::remove_reference<Unwrapped>::type, T>::value &&
+            std::is_base_of<Base, typename std::decay<Unwrapped>::type>::value,
+        int>::type = 0>
+inline auto invoke_impl(T Base::*pmf, Derived&& ref, Args&&... args)
+    FPLUS_INVOKE_RETURN((std::forward<Unwrapped>(ref).*
+                         pmf)(std::forward<Args>(args)...))
+
+// pointer to member function - pointer to object
+template <
+    typename Base,
+    typename T,
+    typename Pointer,
+    typename... Args,
+    typename std::enable_if<
+        is_function<T>::value &&
+            has_correct_cv<typename std::remove_pointer<
+                               typename std::decay<Pointer>::type>::type,
+                           T>::value &&
+            !std::is_base_of<Base, typename std::decay<Pointer>::type>::value,
+        int>::type = 0>
+inline auto invoke_impl(T Base::*pmf, Pointer&& ptr, Args&&... args)
+    FPLUS_INVOKE_RETURN(((*std::forward<Pointer>(ptr)).*
+                         pmf)(std::forward<Args>(args)...))
+
+// pointer to non-static data member - reference to object
+template <
+    typename Base,
+    typename T,
+    typename Derived,
+    typename Unwrapped = unwrap_reference_wrapper_t<Derived>,
+    typename std::enable_if<
+        !is_function<T>::value &&
+            std::is_base_of<Base, typename std::decay<Unwrapped>::type>::value,
+        int>::type = 0>
+inline auto invoke_impl(T Base::*pmd, Derived&& ref)
+    FPLUS_INVOKE_RETURN((std::forward<Unwrapped>(ref).*pmd))
+
+// pointer to non-static data member - pointer to object
+template <
+    typename Base,
+    typename T,
+    typename Pointer,
+    typename std::enable_if<
+        !is_function<T>::value &&
+            !std::is_base_of<Base, typename std::decay<Pointer>::type>::value,
+        int>::type = 0>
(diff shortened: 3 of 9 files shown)


Dobiasd:
Wow! I'm extremely impressed. That looks like a whole lot of work. I will now start to digest the code. :)

theodelrieu:
Thanks! I didn't write a lot of comments though, so feel free to ask, I shall add more when continuing this PR.

Dobiasd:
OK, `meta.hpp` seems straight-forward.

`invoke_test.cpp` looks very well written-out. Having extensive tests is invaluable for such advanced stuff.

I understand you use `has_function_traits` to switch at compile time. Generic lambdas are made possible to use but the other functions keep their static asserts. Very good!

`zip_with_tag` is there to specialize `arity_static_asserts`. Other functions will get other tags later. I feel it could be nice in the future to separate the asserts from the actual functions that way.

I see `invoke_result_t` replaces `std::result_of` in order to be able to also handle our generic lambdas. `invoke_impl` still looks a bit magical to me. I will study it again tomorrow. But especially that you managed to do all this in a C++11-compatible way is very cool.

Did you already have a look at the [travis results](https://travis-ci.org/Dobiasd/FunctionalPlus/builds/268726231)?

theodelrieu:
Right, for now I've put everything into `pairs.hpp`, but that will of course move somewhere else.

I have to move some files around too, `invoke/meta.hpp` belongs in a `detail` folder.
`std::result_of` is broken, and is deprecated in C++17, but implementing it with `invoke_result` is trivial, I will add it, there might be places where we want `detail::result_of`.

`invoke_impl` is really hard to read indeed.

As for Travis, this is simply a matter of `#pragma`, I will commit soon.

theodelrieu:
There seems to be a problem related to libc++, at least that's what I've inferred from my small research.

Dobiasd:
Mhh, do you think we could find a reasonable workaround that prevents us from forcing users to deviate from their default libc++ version?

theodelrieu:
Yep, writing `std::is_function` ourselves. I'll do that later today, I'm already used to copy-paste `const volatile &&` everywhere anyway

theodelrieu:
Adding a C++17 Travis job should be interesting, I had quite a lot of compiler errors like this:

```cpp
namespace fplus::fwd {
auto apply (...)
}

// some_test_file.cpp
using namespace fplus::fwd;

apply(some_func, std::vector<int>{});
```

> error: call to apply is ambiguous (std::apply | fplus::fwd::apply)

In addition to that, we should add a partial specialization for `noexcept` methods if C++17 is detected.

Dobiasd:
Oh, yeah, `using namespace fplus::fwd;` then probably is not that good an idea. We should remove it from `fwd_test.cpp`.

theodelrieu:
I couldn't reproduce the clang 3.9 build errors, it works perfectly fine on Debian 8 clang 3.9...

Dobiasd:
Yes, the [clang tests](https://github.com/Dobiasd/FunctionalPlus/blob/master/.travis.yml#L45) on travis run against libc++.

theodelrieu:
What version of libc++ is on Travis? I installed latest trunk version and it works fine.

Dobiasd:
Travis uses Ubuntu 14.04 (Trusty) and (according to [its output](https://travis-ci.org/Dobiasd/FunctionalPlus/jobs/269001000))

    sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install clang-4.0 libc++-dev

to install.

apt-get then writes

    Get:1 http://us-central1.gce.archive.ubuntu.com/ubuntu trusty/universe amd64 libc++1 amd64 1.0~svn199600-1 [239 kB]
    Get:2 http://us-central1.gce.archive.ubuntu.com/ubuntu trusty/universe amd64 libc++-helpers all 1.0~svn199600-1 [4,798 B]
    Get:3 http://us-central1.gce.archive.ubuntu.com/ubuntu trusty/universe amd64 libc++-dev amd64 1.0~svn199600-1 [553 kB]

does this number (svn199600) help?

Dobiasd:
If I read the raw logs of the travis builds correctly (output of `sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install clang-3.9 libc++-dev`) it should be like this:

    clang-4.0: libstdc++-4.8
    clang-3.9: libstdc++-4.8
    clang-3.8: libstdc++-4.8
    clang-3.7: libstdc++-4.9
    clang-3.6: libstdc++-4.9

However I do not know why the newer clang versions use an older libstdc++ version.

Dobiasd:
The versions of libstdc++ can be found in the raw logs on travis (output of `apt-get`):

    clang-4.0: libstdc++-4.8
    clang-3.9: libstdc++-4.8
    clang-3.8: libstdc++-4.8
    clang-3.7: libstdc++-4.9
    clang-3.6: libstdc++-4.9

I do not know why the newer clang versions have older libstdc++ versions installed

Using this

    strings /usr/lib/x86_64-linux-gnu/libstdc++.so.6 | grep LIBCXX

outputs the following on travis on all clang versions (I do not know if it is helpful).

    GLIBCXX_3.4
    GLIBCXX_3.4.1
    GLIBCXX_3.4.2
    GLIBCXX_3.4.3
    GLIBCXX_3.4.4
    GLIBCXX_3.4.5
    GLIBCXX_3.4.6
    GLIBCXX_3.4.7
    GLIBCXX_3.4.8
    GLIBCXX_3.4.9
    GLIBCXX_3.4.10
    GLIBCXX_3.4.11
    GLIBCXX_3.4.12
    GLIBCXX_3.4.13
    GLIBCXX_3.4.14
    GLIBCXX_3.4.15
    GLIBCXX_3.4.16
    GLIBCXX_3.4.17
    GLIBCXX_3.4.18
    GLIBCXX_3.4.19
    GLIBCXX_3.4.20
    GLIBCXX_3.4.21
    GLIBCXX_3.4.22
    GLIBCXX_3.4.23
    GLIBCXX_3.4.24

However finding out something about libc++ seems to be more difficult.
`apt-get` only tells us a svn number (`svn199600`). Does this help?

launchpad.net seems to [usually provide](https://launchpad.net/ubuntu/+source/libc%2B%2B) version information, but for Trusty there is also just `1.0~svn199600-1`

    Artful Aardvark: 3.9.1-3
    Zesty Zapus: 3.9.1-2
    Xenial Xerus: 3.7.0-1
    Vivid Vervet: 3.5-2
    Trusty Tahr: 1.0~svn199600-1

However, before this becomes too annoying or makes the code more complicated the needed, we should first research if using libc++ on linux is quite normal or if our travis tests are a special case and almost nobody would be hurt by going away from it, especially since the new libc++ version seems to work OK.

Dobiasd:
The versions of libstdc++ can be found in the raw logs on travis (output of `apt-get`):

    clang-4.0: libstdc++-4.8
    clang-3.9: libstdc++-4.8
    clang-3.8: libstdc++-4.8
    clang-3.7: libstdc++-4.9
    clang-3.6: libstdc++-4.9

I do not know why the newer clang versions have older libstdc++ versions installed

Using this

    strings /usr/lib/x86_64-linux-gnu/libstdc++.so.6 | grep LIBCXX

outputs the following on travis on all clang versions (I do not know if it is helpful).

    GLIBCXX_3.4
    GLIBCXX_3.4.1
    GLIBCXX_3.4.2
    GLIBCXX_3.4.3
    GLIBCXX_3.4.4
    GLIBCXX_3.4.5
    GLIBCXX_3.4.6
    GLIBCXX_3.4.7
    GLIBCXX_3.4.8
    GLIBCXX_3.4.9
    GLIBCXX_3.4.10
    GLIBCXX_3.4.11
    GLIBCXX_3.4.12
    GLIBCXX_3.4.13
    GLIBCXX_3.4.14
    GLIBCXX_3.4.15
    GLIBCXX_3.4.16
    GLIBCXX_3.4.17
    GLIBCXX_3.4.18
    GLIBCXX_3.4.19
    GLIBCXX_3.4.20
    GLIBCXX_3.4.21
    GLIBCXX_3.4.22
    GLIBCXX_3.4.23
    GLIBCXX_3.4.24

However finding out something about libc++ seems to be more difficult.
`apt-get` only tells us a svn number (`svn199600`). Does this help?

launchpad.net seems to [usually provide](https://launchpad.net/ubuntu/+source/libc%2B%2B) version information, but for Trusty there is also just `1.0~svn199600-1`

    Artful Aardvark: 3.9.1-3
    Zesty Zapus: 3.9.1-2
    Xenial Xerus: 3.7.0-1
    Vivid Vervet: 3.5-2
    Trusty Tahr: 1.0~svn199600-1

However, before this becomes too annoying or makes the code more complicated the needed, we should first research if using libc++ on linux is quite normal or if our travis tests are a special case and almost nobody would be hurt by going away from it, especially since the new libc++ version seems to work OK.
@offa Do you know more about this?

theodelrieu:
I tried with clang/libc++ 3.9 and it works fine.

theodelrieu:
running `svn ls -v ^/libcxx/tags` shows me that revision 199600 is located between `RELEASE_33` and `RELEASE_34`, the later tag was added one June 18 2014.

Dobiasd:
Good investigation!
I think it is OK if we drop support for this version. In fact I think it should be enough to only support libc++ from version 3.7 on, since this is the version shipped with Ubuntu 16.04.3 LTS (Xenial Xerus). I guess our travis configuration could be adjusted accordingly. It will try to do this later today.

theodelrieu:
Great, hope this will suffice.

Dobiasd:
OK, I committed [a temporary hack](https://github.com/Dobiasd/FunctionalPlus/commit/9a7841af20cf3120c368243edb41d97dee9dd651) that hopefully bring travis to use version 3.7. I guess you have to merge upstream commits into your local branch, so we can see the effects here.

theodelrieu:
Nope, same issue. if only the errors could help. I will try tomorrow with libc++ 3.7

Dobiasd:
I'm not sure if my [hack](https://github.com/Dobiasd/FunctionalPlus/commit/9a7841af20cf3120c368243edb41d97dee9dd651) works and if Travis really used 3.7 then.
But we know that 3.9 generally works with your changes. So I just [changed](https://github.com/Dobiasd/FunctionalPlus/commit/2202ae4e21ab855ec2517865a5c5b3e7e340df51) Travis from 3.7 to 3.9. If it then also does not work, we know that Travis ignores me. If it works, we know that we are not compatible with 3.7 any more.

Dobiasd:
Oh, I think I found out how to do it in a clean way. One can choose the apt packages in `travis.yml`. This allowed me to [match the libc++ version to the clang version](https://github.com/Dobiasd/FunctionalPlus/commit/cb433e8be2925a3ede05278a103c404bc25fe427). So now we will get test results for different libc++ version with one push. I hope this will speed up our feedback loop and help overcome this bottleneck.

Dobiasd:
It did not work. The packages could not be installed. My [next attempt](https://travis-ci.org/Dobiasd/FunctionalPlus/builds/269463565?utm_source=github_status&utm_medium=notification) also failed.

@offa Sorry to bother you again, but I am really stuck here. Could you help me out? I would like the builds with the different clang versions on Travis to use the matching libc++ version.

offa:
This looks weird. Generally the config used in *master* should work. I have no problems locally using Clang 4.0 and GCC 7.

Dobiasd:
The config is master does work, also with the different clang versions. However to me it seems like the libc++ version used is always `svn199600` independent of the clang version.

offa:
Ah, yes. Checked this on some other projects, it's 199600 there too.

offa:
I'm just wondering: If the solution would be a newer version of libc++, does this require clients to do this too?

Dobiasd:
Perhaps, yes. It's surely a tradeoff; generic lambda support vs. compatibility with older libc++ versions.

But I hope we do not have to increase the minimum required version too much. I guess 3.7 would be tolerable, since this is the default version on Ubuntu 16.04 and thus also Linux Mint from version 18 on.

offa:
Yes, clang 3.7(+) is fully acceptable.

offa:
Maybe it's time to switch to C++14 anyway?

Dobiasd:
Offa, I guess you are right. The decision to be C++11 compatible was made some time ago and probably is outdated today. So yeah, let's just do it!
[done](https://github.com/Dobiasd/FunctionalPlus/commit/43469998234f3cced9e0f43503b27f94859bfcfc) :)

theodelrieu, I hope the (now lifted) restriction to C++11 did not cause you too much additional trouble up to now. From now on we can go ahead and use C++14 freely.

theodelrieu:
Glad to hear that! I'll remove C++11 boilerplate and commit again.

theodelrieu:
@offa I've reset my commits prior to my C++11 fix attempts, including your PR, hope you don't mind.

theodelrieu:
I'll get the llvm-3.7 revision and try again

theodelrieu:
I finally succeeded to reproduce the error. I'm on it.

theodelrieu:
I found out the reason why the code breaks, this has nothing to do with `invoke`, but with libc++'s old `__invoke` implementation.

`mem_fn` returns a `Callable`, but internally uses `__invoke` in their call operator.

Unfortunately, version 3.7 did not fix the defect with `reference_wrapper`, hence the breakage.
It has been fixed in version 3.8.1.

Anyway it's not an issue for this library, since we only take function objects. I will remove tests with `mem_fn`.

Dobiasd:
That sounds good. So perhaps we can remain compatible with the default libc++ version on trusty. That would be nice of course.

@offa: In that case we would not have to care about how to use newer libc++ versions on travis any more. From the fact that such a change was not included in your recent commits, I conclude that it is not very easy to implement anyway.

Dobiasd:
Oh, @theodelrieu, does this mean things like the following (taken from the README.md) will no longer be possible because of `mem_fn`?
```cpp
```

theodelrieu:
It will, as long as `cats` is not a `vector<std::reference_wrapper<Cat>>`.

But that is already the case right now on `master`, since this is a libc++ bug, that was fixed in 3.8.1

theodelrieu:
So what is the next step? If I understand correctly, Travis still have the old libc++ (svn rev 199600), and this is why it complains about `std::plus<void>`.

rev 199600 (Not SFINAE-friendly):
```cpp
    auto operator()(_T1&& __t, _T2&& __u) const
        { return _VSTD::forward<_T1>(__t) + _VSTD::forward<_T2>(__u); }
```

Whereas in 3.7:
```cpp
    auto operator()(_T1&& __t, _T2&& __u) const
    _NOEXCEPT_(noexcept(_VSTD::forward<_T1>(__t) + _VSTD::forward<_T2>(__u)))
    -> decltype        (_VSTD::forward<_T1>(__t) + _VSTD::forward<_T2>(__u))
        { return        _VSTD::forward<_T1>(__t) + _VSTD::forward<_T2>(__u); }
```

Dobiasd:
Ah OK. So we do not need 3.8.1 any more but 3.7 is still needed. OK let's see if @offa can do some travis magic. ;-)

In the meantime you could temporary remove `- libc++-dev` from `travis.yml` in your branch to check if the code works with the old clang versions als libstdc++ if you like. Otherwise, how much done are you? Can the PR be merged when the libc++ problem is solved?

theodelrieu:
`invoke` is not done. However, it is largely enough for the library usage.
I need to clean and rebase my commits, put `meta.hpp` and `invoke.hpp` in a `detail` folder, plus moving the `static_assert` mechanism somewhere else.

After that, I think it will be ready to merge.

I wanted to ask you something: Do you mind if I continue to work on `invoke` on my own?
I don't remember seeing an implementation for C++11, and I'd like to provide it as a library on its own, I believe it could be useful to some people, especially library writers who handle `Callable` objects.

offa:
> @offa I've reset my commits prior to my C++11 fix attempts, including your PR, hope you don't mind.

Of course not! 😄  It was only a needed Workaround for C++11, but is no longer needed since 14.

@Dobiasd Unfortunately using libstdc++ wont work on newer clang versions.

@theodelrieu is there' any patch or workaround is there any patch or workaround mentioned?

Dobiasd:
@theodelrieu Having such a free library in the community could be quite valuable. So of course I do not mind. :)

@offa OK, I figured that from the latest travis logs. ;)
Can you already make an educated guess on how much work it would be to get travis do use different libc++ versions? Perhaps this would be a nice test in general, not just for this PR.

theodelrieu:
@offa There is no workaround I can think of, unfortunately. But it might not be a big issue, since the SFINAE dispatch on `! is_invocable` is only done before a `static_assert`, so you would get a big ununderstandable error instead of the `static_assert` message. We can remove the test that fails to compile to solve that.

@Dobiasd Great!

offa:
> Can you already make an educated guess on how much work it would be to get travis do use different libc++ versions? Perhaps this would be a nice test in general, not just for this PR.

If we can get it up with some PPA it should work. But otherwise it'll be tricky, given we should maintain an acceptable build time.

theodelrieu:
@Dobiasd It happens only when using a function/function object that returns `auto`, and has an ill-formed definition (e.g. `is_invocable<std::plus<>{}, int, string>`)

Dobiasd:
OK, I think it would be OK if we only get the default error messages for these cases if this makes using libc++ svn199600 possible again.

theodelrieu:
I'm afraid the libstdc++ version is too old on Travis. 4.8 doesn't support C++14 IIRC

Dobiasd:
> I'm afraid the libstdc++ version is too old on Travis. 4.8 doesn't support C++14 IIRC

I guess I'm a little bit confused right now. It seems to work with clang versions <= 3.7?

The current state should be the following:

    gcc-7 (+libstdc++): working
    gcc-6 (+libstdc++): working
    gcc-5 (+libstdc++): working
    gcc-4.9 (+libstdc++): working
    clang-4.0 (+libstdc++): broken
    clang-3.9 (+libstdc++): broken
    clang-3.8 (+libstdc++): broken
    clang-3.7 (+libstdc++): working
    clang-3.6 (+libstdc++): working
    clang-4.0 (+libc++ svn199600): broken
    clang-3.9 (+libc++ svn199600): broken
    clang-3.8 (+libc++ svn199600): broken
    clang-3.7 (+libc++ svn199600): broken
    clang-3.6 (+libc++ svn199600): broken

I guess if we manage to [install a newer libc++ on travis](https://stackoverflow.com/questions/45946131/using-different-libc-versions-on-travis-for-clang), this would fix the tests, right? I think simply removing the clang tests altogether would not be a good solution in the long term.

theodelrieu:
I forgot that I just had to put back libc++ in the Travis file...

Dobiasd:
OK, no problem. I am sorry that this annoying stuff is blocking the progress. But I hope it will become a lot smoother once we have sorted this out.

theodelrieu:
No worries, it all work now since I removed the infamous `static_assert`

Dobiasd:
Cool, we keep our compatibility even to the old libc++ version. :)
So I guess we are ready to merge now?

theodelrieu:
Yup :)

offa:
Travis has announced an [update of the Trusty Images](https://docs.travis-ci.com/user/build-environment-updates/2017-09-06/) and is includes an update of Clang to 3.9. As this is in a view days (6.9.17), maybe we can check this again

Dobiasd:
@offa, good to know, thanks. I guess it is not urgent any more and we can postpone it. :)

@theodelrieu, thank you very much. I could not have done what you did on my own. Your contribution is not only sophisticated but also implemented cleanly. It improves FunctionalPlus immensely. The ability to support generic lambdas (and transparent function objects) makes the library much more future-proof.

theodelrieu:
Glad to help!

PR #86, opened 2017-08-29
Some CMake modernizations

offa:
Some CMake modernizations and `CPP14TEST` as an option so it's visible to the user.

CMake v3.2 (or newer) is now required. This is the latest version available on Travis atm..


Dobiasd:
Thanks a lot again. :+1: 

I just saw the files resulting from `make unittest` now have a different path as before. The executables are now located in the `test` directory. Some other files too at not catched by `.gitignore` any more. Could you have a look at it and try to let them be created in a location where they produce less clutter, a temp directory or something?

PR #89, opened 2017-09-04
help GCC 7 to disambiguate has_function_traits with std::function

### LINKED ISSUE #88: transform_pair not working with std::function
The following code does not compile any more. I guess it has something to do with PR #87.

```cpp
const std::function<int(int)> double_int = [](int x) -> int
{
    return 2 * x;
};
typedef std::pair<int, int> IntPair;
fplus::transform_pair(double_int, double_int, IntPair({2,3}));
```

@theodelrieu, could you please have a look at it?

theodelrieu:
fixes #88

### DIFF
--- a/include/fplus/function_traits.hpp
+++ b/include/fplus/function_traits.hpp
@@ -329,16 +329,34 @@ namespace fplus
 {
 namespace detail
 {
+template <typename>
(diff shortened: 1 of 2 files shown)


Dobiasd:
With regression test, nice. :+1:

PR #90, opened 2017-09-08
Clang 5.0 Builds and some related fixes

offa:
Clang 5.0 has been released and is already available on Travis. This PR adds a build using the latest clang and fixes two shadowing warnings reported by it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,13 @@ matrix:
           sources: &sources
             - ubuntu-toolchain-r-test
             - llvm-toolchain-precise
-            - llvm-toolchain-precise-3.9
             - llvm-toolchain-precise-3.8
             - llvm-toolchain-precise-3.7
             - llvm-toolchain-precise-3.6
-            - sourceline: 'deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-4.0 main'
-              key_url: 'http://apt.llvm.org/llvm-snapshot.gpg.key'
+            - llvm-toolchain-trusty
+            - llvm-toolchain-trusty-3.9
+            - llvm-toolchain-trusty-4.0
+            - llvm-toolchain-trusty-5.0
     - env: CXX=g++-6 CC=gcc-6
       addons:
         apt:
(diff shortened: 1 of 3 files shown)


offa:
As the list of builds has become considerable long, what about reducing it to eg. GCC 5 - 7 and Clang 3.9 - 5.0?

Dobiasd:
Thanks a lot again. I really appreciate you caring for quite some time now.

offa:
Btw., this PR included the 

# 🎆 1.000 Commit  🎆 #

Dobiasd:
I'm not sure about removing the old compilers.

For me personally the build times on travis did not feel like slowing down my workflow yet. How is your experience?

Also I do not know how many users unable to update their compiler for some reason are out there.

For g++ it seems like version 5.0 is the first one with full C++14 support. For clang it is version 3.4. At least [according to cppreference](http://en.cppreference.com/w/cpp/compiler_support). So wouldn't it make more sense to remove g++ 4.9 but add clang 3.5 and clang 3.4? On the other hand this would feel unbalanced if one just regards the release dates ([clang](http://releases.llvm.org/), [gcc](https://gcc.gnu.org/releases.html)).

offa:
> How is your experience?

< 10 Min. is still ok to me.


> So wouldn't it make more sense to remove g++ 4.9 but add clang 3.5 and clang 3.4? 

I don't think so. Most OS ship newer compiler or at least have them accessible somehow. Even Travis with running on trusty one can easily get some recent compiler. The current Ubuntu LTS (16.04) ships Gcc 5 / Clang 3.8 an example. Those who are already on C++14 (hopefully) have at least one of these two available. But yes, unfortunately, I don't know the compilers used either.

Dobiasd:
OK, I guess you are right.

And I think Ubuntu LTS is a good cornerstone. So for me it would be OK to remove everything before gcc 5 and clang 3.8.

offa:
I'll open an PR, but first I have to check if we can drop the precise APT repository at this point.

PR #92, opened 2017-09-09
Feature/generic lambdas

theodelrieu:
Hi, I managed to upgrade result.hpp, maybe.hpp and a part of composition.hpp.

I reimplemented `compose*` functions to be variadic in the meantime, but I'm not sure of the wording I chose for the implementation functions in `detail`...

Furthermore I have no idea how to update the API Search doc to reflect the variadic-ness of those functions.

Looking forward to your comments.

### DIFF
--- a/include/fplus/compare.hpp
+++ b/include/fplus/compare.hpp
@@ -7,6 +7,8 @@
 #pragma once
 
 #include <fplus/function_traits.hpp>
+#include <fplus/detail/invoke.hpp>
+#include <fplus/detail/asserts/compare.hpp>
 
 namespace fplus
 {
@@ -266,17 +268,16 @@ Y always_arg_2_of_2(const X&, const Y& y)
 // f(x) == g(y)
 // Provides an equality check of two values
 // after applying a transformation function each.
-template <typename F, typename G,
-    typename FIn = typename utils::function_traits<F>::template arg<0>::type,
-    typename GIn = typename utils::function_traits<G>::template arg<0>::type,
-    typename FOut = typename std::result_of<F(FIn)>::type,
-    typename GOut = typename std::result_of<G(GIn)>::type>
-std::function<bool(const FIn& x, const GIn& y)>
-        is_equal_by_and_by(F f, G g)
+template <typename F, typename G>
+auto is_equal_by_and_by(F f, G g)
 {
-    internal::check_compare_preprocessors_for_types<F, G, FIn, GIn>();
-    return [f, g](const FIn& x, const GIn& y)
-    {
+    return [f, g](const auto& x, const auto& y) {
+        (void)detail::trigger_static_asserts<detail::is_equal_by_and_by_tag,
+                                             F,
+                                             decltype(x)>();
+        (void)detail::trigger_static_asserts<detail::is_equal_by_and_by_tag,
+                                             G,
+                                             decltype(y)>();
         return is_equal(f(x), g(y));
     };
 }
--- a/include/fplus/composition.hpp
+++ b/include/fplus/composition.hpp
@@ -7,6 +7,8 @@
 #pragma once
 
 #include <fplus/function_traits.hpp>
+#include <fplus/detail/asserts/composition.hpp>
+#include <fplus/detail/composition.hpp>
 
 #include <functional>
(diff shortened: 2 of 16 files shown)


theodelrieu:
Also, `lift_maybe` could be reimplemented to be variadic too, I did not do it here but it would be great IMO

Dobiasd:
Wow, very nice. I especially like the fact that much of the complexity is moved out of the actual functions and compressed in the generic parts. And monstrosities like the following one simply vanish. :D

```cpp
template <typename F, typename G, typename H, typename I,        +template <typename... Callables>
    typename FIn = typename std::remove_const<typename std::remove_reference<      +auto compose_result(Callables&&... callables)
        typename utils::function_traits<F>::template arg<0>::type>::type>::type,      
    typename FOut = typename std::remove_const<typename std::remove_reference<        
        typename std::result_of<F(FIn)>::type>::type>::type,      
    typename GIn = typename std::remove_const<typename std::remove_reference<     
        typename utils::function_traits<G>::template arg<0>::type>::type>::type,      
    typename GOut = typename std::remove_const<typename std::remove_reference<        
        typename std::result_of<G(GIn)>::type>::type>::type,      
    typename HIn = typename std::remove_const<typename std::remove_reference<     
        typename utils::function_traits<H>::template arg<0>::type>::type>::type,      
    typename HOut = typename std::remove_const<typename std::remove_reference<        
        typename std::result_of<H(HIn)>::type>::type>::type,      
    typename IIn = typename std::remove_const<typename std::remove_reference<     
        typename utils::function_traits<I>::template arg<0>::type>::type>::type,      
    typename IOut = typename std::remove_const<typename std::remove_reference<        
        typename std::result_of<I(IIn)>::type>::type>::type,      
    typename Ok = typename IOut::ok_t,        
    typename Error = typename IOut::error_t>
```
This also makes the results of the API search much nicer to read.

I too have no idea how to represent variadic functions in a meaningful way for the API search. And even if we would agree on a syntax, I don't know how I should adjust the rating algorithm in the [search code](https://github.com/Dobiasd/FunctionalPlus/blob/5ac967e0ed1f57707f9663ef3e0093e396a9241c/api_search/frontend/src/TypeSignature.elm#L412). I find it already quite hard to maintain.
But I also do not like to remove the `API search type` line completely like you did in the case of `compose`, since I would like to keep these functions visible in the API search results. Perhaps we should generelly keep just the most simple case in the type annotation and add a comment that more arguments are possible, for example:
```cpp
// API search type: compose : ((a -> b), (b -> c)) -> (a -> c)
// Forward function composition.
// compose(f, g)(x) = g(f(x))
// It is also possible to compose more than two functions:
// compose(f, g, h)(x) = g(h(f(x)))
```

I like the fact that you sprinkled the tests with `std::plus<>`. :D

And I do not see a wording problem in `details`.

theodelrieu:
Oh, didn't intend to delete one, I might have a swift delete-button finger.

I too believe that adding a comment is good enough. So `bind_maybe`, `binary_lift` are correct? I'll make the changes tonight.

Dobiasd:
Yes, they look good to me. I also just tested them with the following chain and it worked fine.

```cpp
    const std::function<fplus::maybe<int>(const std::string&, int, double)>
        string_to_maybe_int =
            [](const std::string& str, int, double) -> fplus::maybe<int> {
        return fplus::just<int>(static_cast<int>(str.size()));
    };

    const auto maybe_int_to_double =
            [](int x) -> fplus::maybe<double> {
        return fplus::just<double>(x + 1.7);
    };

    struct maybe_double_to_int {
        fplus::maybe<int> operator()(double x) const {
            return fplus::just<int>(fplus::round(x));
        }
    };

    REQUIRE_EQ(compose_maybe(
        string_to_maybe_int,
        maybe_int_to_double,
        maybe_double_to_int())(std::string("hi"), 2, 3.4), just(4));
```

theodelrieu:
I put back the API Search comment, and modified every `compose` function I changed.
About the wording, I've replaced `functions` with `callables`, I can revert that if you want.

Dobiasd:
`callables` is OK with me, since it actually it is more true than `functions`. :)

PR #94, opened 2017-09-14
Fix master

theodelrieu:
Reverts every commit related to generic lambda support. They will be moved into a feature branch right after.

This should be merged only once the work is 100% done, otherwise some combination of functions will fail to compile.


theodelrieu:
Well, it seems libc++ strikes again, the error is triggered by this line:

```cpp
```

The `mem_fn` implementation uses `__invoke` internally, and checks for `std::is_function` at some point. Just like the issue I encountered earlier when implementing `invoke`...

I'm afraid there is only one solution, and that's upgrading libc++ on Travis.
The version we're using isn't an official release, according to the svn revision (I wonder why Travis is using it therefore).

Dobiasd:
I had no luck with my [question on SO](https://stackoverflow.com/questions/45946131/using-different-libc-versions-on-travis-for-clang?noredirect=1#comment79207241_45946131) regarding newer libc++ version on travis, despite 150 points as a bounty. ;-)

I even failed to install libc++ 3.7 on a local trusty virtualbox. I guess one has to build it from source. However I am not sure if this is the right solution to use on travis. Thus I suggest we simply remove libc++ for now from our `travis.yml`.

theodelrieu:
Fair enough, I believe we should add macOS to Travis, as well as AppVeyor for Windows.

Dobiasd:
Great idea! Do you have any experience with this?

theodelrieu:
Absolutely not! That will be interesting anyway :)

Dobiasd:
Clang <= 3.9 has a problem with `::gets`. [According to SO](https://stackoverflow.com/questions/24342312/clang-seems-to-use-the-gcc-libraries) using libc++ with clang solves it. -_-

theodelrieu:
Let's hope...

Dobiasd:
Just started a new attempt to install libc++ on trusty, and opened a [new question](https://stackoverflow.com/questions/46232955/how-to-install-libc-3-9-on-ubuntu-trusty-without-building-from-source). Let's see what happens.

theodelrieu:
Yes I think we should merge it

Dobiasd:
OK, and I re-enables libc++ for now, because of the problem with libstdc++ and old clang versions. At least our tests are green one again that way. ;)

In your feature branch you can of course kick it out again and also delete the clang versions not working with libstdc++, so you have a more friendly test feedback. When merging at some point we can re-evaluate our options.

theodelrieu:
Great, I'll try to add macOS and Windows for tests this weekend.

Dobiasd:
Oh, it looks like I cheered too soon. The clang tests are still red on master.

theodelrieu:
Yes, that's the issue:

Leaving the old libc++ breaks because of a bug related to `std::is_function`.
Removing it breaks because of `::gets`...

theodelrieu:
A workaround would be to remove the `std::mem_fn` test, and wait for multi-platform CI

theodelrieu:
I've pushed on branch `feature/generic_lambdas`. Sorry for the mess!

PR #95, opened 2017-09-15
add xcode8 to travis. refactor travis file to not use libc++ on linux

theodelrieu:
Here is an attempt to fix the CI on unix systems.

Adding xcode8 support, and removing libc++-dev on Linux.


theodelrieu:
I looked at the travis file of this [JSON](https://github.com/nlohmann/json) library for this PR.

offa:
I don't think OS X is a good idea here. Those builds typically increase the build time dramatically (have even encountered > 2 h) and have problems and downtimes very often.

Dobiasd:
I just started to add ugly `#if !defined _LIBCPP_VERSION || (_LIBCPP_VERSION != 1101)` around the tests using `std::men_fn`, but your solution on fixing our CI is a lot better of course. Nice!

Travis seems to not run the osx tests without at least one `env`. The json lib you linked has something global in [its configuration](https://github.com/nlohmann/json/blob/e45eaf6e3085fd5e51907cff1b93d747620ffc00/.travis.yml#L17).

theodelrieu:
Oh right, I didn't see that!

Dobiasd:
I'm not sure it `no environment variables set` is the reason for the build not starting, or if it is just a remark and we just would have to wait. But if the two hours thing turns out to be true, I would consider this way too long.

theodelrieu:
It is just a remark, here the VM hasn't booted yet

theodelrieu:
We could keep a commit that adds all macOS images to the Travis file on a separate branch,
and just before a new release trigger a PR with this only commit.

So that we only check macOS once per release (more if there is a breakage at that point obviously)

theodelrieu:
I will open a new PR (mainly because of the branch name)

Dobiasd:
OK

PR #100, opened 2017-09-25
Feature/generic lambdas

theodelrieu:
Hi, I converted `composition.hpp`, with the exception of `memoize` functions.


theodelrieu:
Oops, I forgot to rebase

theodelrieu:
I forgot to fetch before making changes, therefore there is only a single commit in this PR.

If the changes are good to you, could you select `Rebase and Merge`? I'd like to keep this option for this branch, it's easier for me to work with than with merge commits, especially if I need to make a sneaky change in a past commit :)

PR #102, opened 2017-09-25
Feature/generic lambdas

theodelrieu:
Converted some `transform` methods. There are not many left :)


theodelrieu:
I forgot to push some commits from a few days ago, I removed the `-Wshadow` warning, which was causing lots of noise on GCC 6 specifically.

Basically it complained with the following code:

```cpp
auto f() {
  return [](auto x){};
}

int main() {
  auto x  = f(); // says that x shadows the parameter of the lambda, but it doesn't.
}
```

PR #110, opened 2017-10-09
Refactor/remove result of

### LINKED ISSUE #106: Clean-up: remove result_of occurences
`result_of` is deprecated in C++17, and has a lot of caveats.

We should use `invoke_result_t` instead. Note that `invoke_result_t` returns a `T&&` (as `std::declval`).

It is important to be careful when storing it into a template type:

```cpp
using Result = detail::invoke_result_t<F>;
// std::vector<T&&> will fail to compile
// using ContainerOut = std::vector<Result>;
using ContainerOut = std::vector<std::decay_t<Result>>;
```

theodelrieu:
Fixes #106


Dobiasd:
Looks good. I will handle the error with the clang 5 set on travis tomorrow.

PR #112, opened 2017-12-05
Cmake improvements

offa:
Some CMake improvements.

- The library is now an interface target (`fplus`) 
- Improved install target
- Modernizations

I used `fplus` as target name. It's shorter than the full name. Indeed it's not consistent with the name used with `project()`. So @Dobiasd please feel free to rename the target to what fits best.


theodelrieu:
Looks good to me.

While you are modernizing it, what do you think about removing `add_compile_options`?

I think it should be replaced with `target_compile_options`, on test targets only

offa:
That's a good point. But, what about the other two targets (`readme_…` and `99_problems`)?

offa:
Does adding those flags as interface options to `fplus` make sense? This would also propagate them to the tests and the two special targets mentioned before.

theodelrieu:
That would also leak to the users of fplus. Interface compile options should only contain **mandatory** flags.

Maybe making those PRIVATE to the target `fplus` is sufficient.

offa:
> That would also leak to the users of fplus. Interface compile options should only contain mandatory flags.

That's true. Private isn't possible as the `fplus` is an interface target (only interface flags are possible).

So I guess the best solution is the one you recommended first. Just one thing: Should the two executable also get those flags?

theodelrieu:
Ah right, well I guess we should add those to both targets indeed.

PR #115, opened 2017-12-18
Allow use `std::locale` in `to_upper_case` and `to_lower_case`

Artalus:
I recently encountered a problem while changing string case in Russian. Quick googling showed that it is possible to easily upgrade FPlus to support `std::locale` which can be used in `toupper` and `tolower`. The only problem - on Windows I couldn't find a way to properly encode UTF-8, it looks like there is no such locale there, but I may be worng. I tried my best to supply tests and fwd + fwd-flip, but the generator script didn't work for me and I had to rewrite include files manually.


Artalus:
The current failure at `result_test.cpp:139` is due to compiler being unable to resolve `to_upper_case<std::string>` as a call to first overload:

    template<class String> String to_upper_case(const String& s);
    template<class String> String to_upper_case(const String& s, const locale replacement);

It can be fixed by either renaming the second overload to something like `to_upper_case_loc` or by passing `[](auto &&x){ return to_upper_case(x); }`

Dobiasd:
Hi and thanks for the contribution. It looks good.

I would prefer to append `_loc` to the function names, since I try to avoid overloads where possible in this library. Could you add this to your commits so we can see the tests pass before merging?

Dobiasd:
Also I think you can swap the parameters of the new functions. I guess they will very often get partially applied, and I think the default case should rest in `fplus::fwd` and not in `fplus::fwd::flip`.

Artalus:
It looks like Clang doesn't like "unknown" encoding strings in source code and fails due to `-Werror` flag; can be probably fixed by adding `-Wno-invalid-source-encoding` to CMakeLists.txt.
GCC compiles it fine though, and test fails due to lack of Russian locale which I used for tests on the CI server. Can probably switch to some other language, but it looks like this test is highly dependant on the environment.

Dobiasd:
Do you have any idea how we could make the tests for the feature more stable? I would like to include it, but not at the cost of making the unit tests fragile.

Artalus:
I believe Travis provide some sort of system pre-configuration, that will allow installation of needed locales. I will conduct some experiments later.

IMO it also wouldn't be wise to _strictly fail_ tests due to locale missing - we can't guarantee that the code isn't working, only that we lack some part of system that user might not even need - but user probably should know that _something went wrong_. No idea how to get such behavior yet, though. Maybe some CMake option to disable unneeded tests...

Dobiasd:
There is also another option. I know that it is "ethically" questionable, but since the feature is quite small, we could perhaps just leave the unit tests out for it. In my opinion testing is just a tool and not a law, so it would be OK with me. :)

If we find a better solution later, we can still include it then. What do you think of this?

Artalus:
I know that on my ArchLinux with installed CP-1251 locale all tests are OK, and that on Windows only UTF-8 tests fail, which seems unfixable. If you would accept the feature without proper unit-tests for it - I won't argue, but still would like to spend some time checking if I can achieve those :)

Dobiasd:
OK, I will merge as soon as the unit tests pass. You can decide if you want to remove them for this and add a stable version of them later in a new PR, or if you want to fix them first before merging this PR.

Artalus:
I added tests for Greek strings (for diversity sake) and made Travis install missing locales so the tests run correctly on CI. However, depending on user OS configuration the same tests may fail due to exception thrown at locale acquiring. For such cases I've added a more graceful way of failing while indicating the problematic locale.

Dobiasd:
Oh, nice. Looks good.

I cloned the PR onto my local machine, but was unable to get the unit tests running. Do you have an idea what I am missing?

```bash
$ lsb_release -a
No LSB modules are available.
Distributor ID: LinuxMint
Description:    Linux Mint 18.2 Sonya
Release:    18.2
Codename:   sonya

$ sudo localedef -c -i ru_RU -f CP1251 ru_RU.CP1251

$ sudo localedef -c -i el_GR -f CP1253 el_GR.CP1253

$ locale -a | grep ru_RU.cp1251
ru_RU.cp1251

$ locale -a | grep el_GR.cp1253
el_GR.cp1253

$ sudo dpkg-reconfigure locales
# confirm in command line GUI

$ shutdown -r now
# just to make sure ;)

$ make unittest

# ...

/home/tobias/Desktop/FunctionalPlus/test/stringtools_cp1251_test.cpp(13)
TEST CASE:  stringtools_test, to_lower/upper_case, cp1251

/home/tobias/Desktop/FunctionalPlus/test/get_locale.hpp(10) FATAL ERROR!
  Couldn't acquire locale: locale::facet::_S_create_c_locale name not valid. Is 'ru_RU.cp1251' supported on your system?
```

Artalus:
This is rather strange. From `_S_create_c_locale` I guess you are compiling with Clang, does GCC gives the same error as well? I can theoretize that on some Linux distributions one may need not only to set locale, but to install some sort of language package or smth like that... I'll try and check this on some other distros that I might encounter in my `Downloads` folder.

Dobiasd:
It's `gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.5)`.

Dobiasd:
I managed to get it running with the following commands:
```bash
$ sudo locale-gen ru_RU
$ sudo locale-gen ru_RU.UTF-8
$ sudo locale-gen el_GR
$ sudo locale-gen el_GR.UTF-8
$ sudo update-locale
```
Is the PR ready to merge from your point of view too?

Artalus:
Good to hear! The only thing that I might consider, is to maybe make `get_locale` somehow just warn user without strictly failing the test, but I am not sure if it can be done with Doctest, neither if that would be a viable option.

Dobiasd:
A warning instead of an error would be nice for users building the unit tests locally without having the languages installed. However this could mean that we could miss real problems with tests on travis. So I think an error is better.

Some minor things I just noticed:
- `COMMAND stringtools_cp1253_test.cpp` is missing in `test/CMakeLists.txt`
- In `test/stringtools_cp1253_test.cpp` it says `TEST_CASE("stringtools_test, to_lower/upper_case, cp1251")` instead of `cp1253`

Could you please correct these? :)

Dobiasd:
In the end: Thanks a lot for this nice new feature. :+1:

PR #116, opened 2018-02-04
add relocatable cmake generated pkgconfig installation

headupinclouds:
... and related changes.

* rename UNITTEST option to FPLUS_BUILD_UNITTEST to avoid name collisions with direct`add_subdirectory()` usage
* add FPLUS_BUILD_{EXAMPLES,UNITTEST} option to support cmake pkconfig build + installation without the tests and examples (default ON based on previous behavior)
* add FPLUS_USE_TOOLCHAIN to allow the user to provide their own toolchain definitions (higher language standard or warnings), which is friendly to package managers (default OFF to mirror current behavior)
* remove `$<INSTALL_INTERFACE:include>` from target_include_diretories (not required)
* use list APPEND for CMAKE_MODULE_PATH to avoid clobbering existing CMAKE_MODULE_PATH definition, and prefer CMAKE_CURRENT_LIST_DIR to CMAKE_CURRENT_SOURCE_DIR
* add INTERFACE linking of Threads::Threads to lib, and include in pkgconfig dependencies
* add cmake generated relocatable pkgconfig installation (CONFIG mode FunctionalPlus package with fplus INTERFACE library) using tested boilerplate from https://github.com/forexample/package-example (application of https://cmake.org/cmake/help/v3.10/module/CMakePackageConfigHelpers.html)

The installed package supports use of FunctionalPlus with syntax like this:

```
find_package(FunctionalPlus CONFIG REQUIRED)

add_executable(foo foo.cpp)
target_link_libraries(foo FunctionalPlus::fplus)
```

NOTE: This enables pkgconfig usage of frugally-deep.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,7 +102,7 @@ before_install:
 
 install:
   - mkdir -p build && cd build
-  - cmake .. -DUNITTEST=ON
+  - cmake .. -DFPLUS_BUILD_UNITTEST=ON
 
 script:
   - which $CXX
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,56 +1,128 @@
 cmake_minimum_required(VERSION 3.2)
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
+
 project(FunctionalPlus VERSION 0.2)
 
 message(STATUS "===( ${PROJECT_NAME} ${PROJECT_VERSION} )===")
 
-option(UNITTEST "Build unit tests" OFF)
+option(FPLUS_USE_TOOLCHAIN "Use compiler flags from an external toolchain" OFF)
+option(FPLUS_BUILD_EXAMPLES "Build examples" ON)
+option(FPLUS_BUILD_UNITTEST "Build unit tests" ON)
 message(STATUS "Building Unit Tests ${UNITTEST}")
 
+if(NOT FPLUS_USE_TOOLCHAIN)
+  set(COMPILE_OPTIONS -Wall
+    -Wextra
+    -pedantic
+    -Werror
+    -Weffc++
+    -Wconversion
+    -Wsign-conversion
+    -Wctor-dtor-privacy
+    -Wreorder
+    -Wold-style-cast
+    -Wparentheses
+    )
 
-set(COMPILE_OPTIONS -Wall
-                    -Wextra
-                    -pedantic
-                    -Werror
-                    -Weffc++
-                    -Wconversion
-                    -Wsign-conversion
-                    -Wctor-dtor-privacy
-                    -Wreorder
-                    -Wold-style-cast
-                    -Wparentheses
-                    )
-
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
-
-
-find_package(Threads REQUIRED)
-
-set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+  set(CMAKE_CXX_EXTENSIONS OFF)
+endif()
 
 add_library(fplus INTERFACE)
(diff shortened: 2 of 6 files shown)


headupinclouds:
These changes seemed to be an atomic commit, and, if accepted, I'm interested in a follow up for frugally-deep.  FunctionalPlus + frugally-deep looks really useful.  Thanks!

Dobiasd:
Thank you very much. I appreciate your contribution a lot, I'd just like to ask fplus's cmake specialist about his opinion before merging. ;)

@offa: What do you think?

offa:
> add FPLUS_BUILD_{EXAMPLES,UNITTEST} option to support cmake pkconfig build + installation without the tests and examples (default ON based on previous behavior)

Now as you say it, an option for the examples is still missing …

> add FPLUS_USE_TOOLCHAIN to allow the user to provide their own toolchain definitions (higher language standard or warnings), which is friendly to package managers (default OFF to mirror current behavior)

Just to get this right: If the option is set of `OFF` the user is responsible the those flags. This includes the correct C++ standard (C++14 atm.). If the user sets the wrong standard or looses some (probably useful) warnings it's fully his fault.

However, the documentation clearly says what standard is required, the option describes well what it does and it's still optional behaviour. Finally, the typical users aren't affected, but some more specialized use cases (eg. packing) are covered too now. Great!


👍 Great contribution towards modern CMake integration.

-------------------------------------------------

Just one thing: The `CMakeLists.txt` consists of large parts related to installation now. I'd suggest to extract this into a dedicated file to clean up the top-level `CMakeLists.txt` a bit.

offa:
There are some usages of the renamed `UNITTEST` option missing:

- https://github.com/headupinclouds/FunctionalPlus/blob/pr.pkgconfig/CMakeLists.txt#L12
- https://github.com/headupinclouds/FunctionalPlus/blob/pr.pkgconfig/.travis.yml#L105
- https://github.com/headupinclouds/FunctionalPlus/blob/master/README.md#way-1-using-cmake (cant link to the line, so please look for `cmake -DUNITTEST=ON ..` there)

headupinclouds:
> There are some usages of the renamed UNITTEST option left:

Thanks for the review and the catches.  I'll send a fix.

> Now as you say it, an option for the examples is still missing …

The option is here:

https://github.com/headupinclouds/FunctionalPlus/blob/db8db8ad266f4805c35b7707c061c7295cc7f8c4/CMakeLists.txt#L10

It is used here:

https://github.com/headupinclouds/FunctionalPlus/blob/db8db8ad266f4805c35b7707c061c7295cc7f8c4/CMakeLists.txt#L46

Let me know if I'm missing something.

> Just to get this right: If the option is set of OFF the user is responsible the those flags. This includes the correct C++ standard (C++14 atm.). If the user sets the wrong standard or looses some (probably useful) warnings it's fully his fault.

Yes, that's the idea.  The default option should preserve current behavior, but will allow the user to override these settings with a central toolchain in the context of multi-package builds.

> Great contribution towards modern CMake integration.

Glad I could help.  Thanks for the project.  I'll put a similar PR together for frugally-deep and look for a review there.

> The CMakeLists.txt consists of large parts related to installation now. I'd suggest to extract this into a dedicated file to clean up the top-level CMakeLists.txt a bit.

I'll move this into something like `cmake/pkgconfig-install.cmake` and `include()` it from the top CMakeLists.txt.

offa:
> The option is here:

What I mean is: We haven't included it yet (prior to this PR) 😄 

> I'll move this into something like cmake/pkgconfig-install.cmake and include() it from the top CMakeLists.txt.

Perfect! 👍

headupinclouds:
After we finalize the merge here, Would you consider hosting tagged + versioned releases (w/ tags reflecting current semantic versioning)?

i.e.,:

https://github.com/Dobiasd/FunctionalPlus/archive/v0.2.tar.gz
https://github.com/Dobiasd/FunctionalPlus/archive/v0.3.tar.gz
https://github.com/Dobiasd/FunctionalPlus/archive/v0.4.tar.gz

Similar to:

https://github.com/headupinclouds/FunctionalPlus/releases

This would be friendly to package managers.

https://github.com/headupinclouds/FunctionalPlus/archive/v0.2-p0.tar.gz

Dobiasd:
Thanks a lot you two guys!

I just added a release and will add new ones in the future if worthwhile things change. Let's see if we ever reach v1.0. ;)

PR #124, opened 2018-05-04
Clang 6 build fixed

offa:
The current Clang 6 is broken due a failing libc++ installation. The download fails because the version of the compiler is used for the filename (that is 6.0.1) which is not available.

Using v6 as trunk version fixes the problem.


Dobiasd:
Thanks so much for caring! :heart:

PR #128, opened 2018-06-14
update `append`, add `transform_append`

scinart:
`append` now use 3 template parameters to allow different container types as input or output parameters.
`transform_append` is same as `append`, but easier to use different result container type.

### DIFF
--- a/include/fplus/container_common.hpp
+++ b/include/fplus/container_common.hpp
@@ -1130,16 +1130,27 @@ ContainerOut prepend_elem(const T& y, Container&& xs)
 // fwd bind count: 1
 // Concatenaes two sequences.
 // append([1, 2], [3, 4, 5]) == [1, 2, 3, 4, 5]
-template <typename Container>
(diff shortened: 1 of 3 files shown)


Dobiasd:
Nice commit, having unit tests, correct docs, etc. :+1: 

I also like that `append` is now more general without breaking the API.

Just one thing regarding the naming of the new function `transform_append`: I see you took `transform_convert` ("Same as transform, but ...") as an example to build it. In `fplus` the verb `transform` is usually used to mean something similar to `std::transform`, and `convert` is used to indicate a different element/container type. So I suggest changing the name `transform_append` to `append_convert` and we can merge your PR. :)

We can ignore the Clang 6.0 error on Travis, since it has the [same problem](https://github.com/Dobiasd/FunctionalPlus/issues/129) on the master branch, that was working fine 6 days ago when the last commit was made.

Dobiasd:
Good contribution! Thank you. :+1:

Issue #129, opened 2018-06-14
Clang 6.0 on travis:   The C++ compiler "/usr/bin/clang++-6.0" is not able to compile a simple test program.


Dobiasd:
https://travis-ci.org/Dobiasd/FunctionalPlus/jobs/389830560

https://api.travis-ci.org/v3/job/389830560/log.txt

It was working fine 6 days ago when I made the last commit. I only broke when I just triggered a rebuild.

offa:
Strange, according to the log, libc++ wasn't found:

>   /usr/bin/ld: cannot find -lc++

offa:
@Dobiasd i'm working on a solution …

Dobiasd:
You are my hero! :metal:

PR #130, opened 2018-06-14
CI Installation of libc++ fixed

offa:
CI Installation of libc++ fixed (#129).


Dobiasd:
Oh, that is a very nice fix. I like the overall complexity going down also. :slightly_smiling_face: 
Thanks a lot! :heart:

PR #135, opened 2018-08-02
compare.hpp : replace [](auto x, decltype(x) y) by static_assert

pthom:
see https://stackoverflow.com/questions/51641219/msvc-lambda-with-two-auto-params-of-same-type#51641584


Dobiasd:
Cool, thanks. :)

PR #136, opened 2018-08-02
Win port

pthom:
# Status / port to windows

Tested under Visual Studio 2015 / 32 bits and 64 bits.

* All projects compile in debug and release mode
* All test pass in 32 and 64 bits
* `cmake install` passes
* Continuous integration with appveyor, using msvc 2015 and 2017 (64 bits)
  See example here: https://ci.appveyor.com/project/pthom/functionalplus

Some tests had to be disabled although; see details below.

## msvc toolchain

A specific toochain was added (`toolchain_msvc.cmake`) : 

Warning level is set to a high level (/W4), and some additional warning were enabled.

Some warnings had to be silenced:
* 4459 # declaration of 'xs' hides global declaration
* 4127 # conditional expression is constant
* 4100 # unreferenced formal parameter
* 4189 # local variable is initialized but not referenced

## invoke_test.cpp

Quite some tests had to be disabled here. I do not know if they would pass with msvc 2017.

I added a macro in order to disable some tests

````
#ifdef _MSC_VER
#define FPLUS_MSVC_BYPASS_FAILING_TESTS
#endif
````

The error types are:

* Error C2338 under MSVC (i.e static_assert fail)
* Error C2672: 'fplus::internal::invoke': no matching overloaded function found

The code behind `fplus::internal::invoke` is really hard. I do not feel like I can solve those.
Help would be appreciated.

## stringtools_utf8_test.cpp

This test is disabled since the locale ru_RU.utf8 is not available.

### DIFF
--- a/test/udemy_course_test.cpp
+++ b/test/udemy_course_test.cpp
@@ -87,7 +87,7 @@ TEST_CASE("udemy_course_test, Programming_challenge_longest_edge_of_polygon")
     using namespace Programming_challenge_longest_edge_of_polygon;
 
     vector<point> polygon =
-        { {1,2}, {7,3}, {6,5}, {4,4}, {2,9} };
+        { {1.f,2.f}, {7.f,3.f}, {6.f,5.f}, {4.f,4.f}, {2.f,9.f} };
 
     const auto edges =
         fplus::overlapping_pairs_cyclic(polygon);
@@ -163,7 +163,7 @@ TEST_CASE("udemy_course_test, High_level_expressiveness_and_concise_code")
     using namespace High_level_expressiveness_and_concise_code;
 
     vector<point> polygon =
-        { {1,2}, {7,3}, {6,5}, {4,4}, {2,9} };
+        { {1.f,2.f}, {7.f,3.f}, {6.f,5.f}, {4.f,4.f}, {2.f,9.f} };
 
     const auto result = fplus::maximum_on(
         edge_length,
@@ -234,7 +234,7 @@ TEST_CASE("udemy_course_test, Forward_application")
     using namespace Forward_application;
 
     vector<point> polygon =
-        { {1,2}, {7,3}, {6,5}, {4,4}, {2,9} };
+        { {1.f,2.f}, {7.f,3.f}, {6.f,5.f}, {4.f,4.f}, {2.f,9.f} };
 
     // 1:
     const auto result = fplus::fwd::apply(polygon
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .vscode
 test/temp*
 build
+build64
 api_search/frontend/build
 api_search/frontend/elm-stuff
 api_search/frontend/deploy.sh
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,11 @@ message(STATUS "Building Unit Tests ${FPLUS_BUILD_UNITTEST}")
 message(STATUS "Building examples ${FPLUS_BUILD_EXAMPLES}")
 
 if(NOT FPLUS_USE_TOOLCHAIN)
-  include(cmake/toolchain.cmake)
+  if (MSVC)
+    include(cmake/toolchain_msvc.cmake)
+  else()
+    include(cmake/toolchain.cmake)
+  endif()
 endif()
 
 add_library(fplus INTERFACE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ message(STATUS "===( ${PROJECT_NAME} ${PROJECT_VERSION} )===")
 option(FPLUS_USE_TOOLCHAIN "Use compiler flags from an external toolchain" OFF)
 option(FPLUS_BUILD_EXAMPLES "Build examples" ON)
 option(FPLUS_BUILD_UNITTEST "Build unit tests" OFF)
+
 message(STATUS "Building Unit Tests ${FPLUS_BUILD_UNITTEST}")
 message(STATUS "Building examples ${FPLUS_BUILD_EXAMPLES}")
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,51 @@
+version: '{build}'
+branches:
+  only:
+  - master
+clone_folder: c:\projects\fplus
+image:
+- Visual Studio 2015
+- Visual Studio 2017
+configuration:
+- Release
+platform:
+- x64
+environment:
+  matrix:
+  - arch: Win64
+  # - arch: #does not work, Release|x64 not a valid target
+matrix:
+  fast_finish: true
+
+# skip unsupported combinations
+init:
+- set arch=
+- if "%arch%"=="Win64" ( set arch= Win64)
+- echo %arch%
+- echo %APPVEYOR_BUILD_WORKER_IMAGE%
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2017" ( set generator="Visual Studio 15 2017%arch%" )
+- if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" ( set generator="Visual Studio 14 2015%arch%" )
+- echo %generator%
+
+before_build:
+- cmd: |-
+    cd c:\projects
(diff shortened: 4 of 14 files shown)

### CODE REVIEW

File: test/udemy_course_test.cpp [hunk 1]
```diff
@@ -14,6 +14,10 @@
 #include <fplus/fplus.hpp>
 #include <vector>
 
+#ifdef _MSC_VER
+#pragma warning(disable: 4127) // conditional expression is constant
+#endif
+
```
pthom: This can probably be removed

pthom:
Continuous integration (with appveyor) us up !
It works under msvc 2015 & 2017, in 64 bits (I could not activate the 32 bits builds, although it works)

(travis does not provide windows targets)

Dobiasd:
Awesome! So is this ready to be merged?

pthom:
Hello,

I am on holiday right now, so that I do not have a lot of time.

But I would like to make a more complete analysis of the failure on `ìnvoke_test.cpp`.

So, I prepared some more details for you, and I would like your opinion.
I suspect the answer might be that the compiler is not 100% compliant and that we should go on as it is, but I am not really sure.

### Analysis under windows 64 bits

Under windows 64 bits, there is actually only one failure (see my last commit), this is a good news!

````cpp
TEST_CASE("member function - derived object reference")
{
  ...
  // this static_assert fails under win64
  static_assert(all_invocable<decltype(qualifiers), derived_function_object_t, int const&, double>::value, "");
 // although this one would pass (not using the derived object)
  static_assert(all_invocable<decltype(qualifiers), function_object_t, int const&, double>::value, "");
````

With a more thorough analysis, I saw that the static assert is due to a failure in the lines I commented out below (i.e if they are commented out, everything compiles correctly)
````cpp
template <typename... TupleArgs, typename Class, typename... FuncArgs>
struct all_invocable<std::tuple<TupleArgs...>, Class, FuncArgs...>
{
  ...
  static constexpr bool value = internal::conjunction<
      internal::is_invocable<Elem<0, Tuple>, Class&, FuncArgs...>,
      internal::is_invocable<Elem<2, Tuple>, Class&, FuncArgs...>,
      internal::is_invocable<Elem<4, Tuple>, Class&, FuncArgs...>,
      internal::is_invocable<Elem<6, Tuple>, Class&, FuncArgs...>
      //internal::is_invocable<Elem<1, Tuple>, Class, FuncArgs...>,
      //internal::is_invocable<Elem<3, Tuple>, Class, FuncArgs...>,
      //internal::is_invocable<Elem<5, Tuple>, Class, FuncArgs...>,
      //internal::is_invocable<Elem<7, Tuple>, Class, FuncArgs...>
  >::value;
};
}
````

Do you have any idea why this might fail? This is strange, because the failure only occurs with the derived class. 

### Analysis under windows 32 bits

There are lots of lines that fail to compile under windows 32 bits. 

#### Example 1: static assert failures
````cpp
TEST_CASE("member function - object reference")
{
   ...
  // this static assert fails
  static_assert(all_invocable<decltype(qualifiers), function_object_t, int const&, double>::value, "");
}
````

With a more thorough analysis, I saw that none of the static asserts inside internal::conjunction below passes
````cpp
template <typename... TupleArgs, typename Class, typename... FuncArgs>
struct all_invocable<std::tuple<TupleArgs...>, Class, FuncArgs...>
{
  ...
  static constexpr bool value = internal::conjunction<
      internal::is_invocable<Elem<0, Tuple>, Class&, FuncArgs...>,
      internal::is_invocable<Elem<1, Tuple>, Class, FuncArgs...>,
      internal::is_invocable<Elem<2, Tuple>, Class&, FuncArgs...>,
      internal::is_invocable<Elem<3, Tuple>, Class, FuncArgs...>,
      internal::is_invocable<Elem<4, Tuple>, Class&, FuncArgs...>,
      internal::is_invocable<Elem<5, Tuple>, Class, FuncArgs...>,
      internal::is_invocable<Elem<6, Tuple>, Class&, FuncArgs...>,
      internal::is_invocable<Elem<7, Tuple>, Class, FuncArgs...>
  >::value;

````

#### Example 2 : Failure to specialize function template

````cpp
TEST_CASE("member function - object pointer")
{
  ...
  auto adder = function_object_t{};
  REQUIRE_EQ(internal::invoke(&function_object_t::operator(), &adder, 40, 2), 42);
}
````

fails with the following message:

````
1>FunctionalPlus\test\invoke_test.cpp(299): error C2672: 'fplus::internal::invoke': no matching overloaded function found
1>FunctionalPlus\test\invoke_test.cpp(299): error C2893: Failed to specialize function template 'invoke_result<F,ArgTypes...>::type fplus::internal::invoke(F &&,ArgTypes &&...)'
1>  FunctionalPlus\test\invoke_test.cpp(299): note: With the following template arguments:
1>  FunctionalPlus\test\invoke_test.cpp(299): note: 'F=int (__thiscall `anonymous-namespace'::function_object_t::* )(int,int) const'
1>  FunctionalPlus\test\invoke_test.cpp(299): note: 'ArgTypes={`anonymous-namespace'::function_object_t *, int, int}'
1>FunctionalPlus\test\invoke_test.cpp(299): error C2672: 'doctest::detail::ResultBuilder::binary_assert': no matching overloaded function found
````

What is your opinion? 



PS : the travis build failure is due to something else (failure to install g++ on some platforms)

Dobiasd:
Thanks a lot for the good analysis!

I guess if we could get the latest MSVC version (i.e., 2017 with the latest updates) to work, this would be OK. I'm not sure about 32 bit, and I had not expected it to make a difference. In general our compatibility goal may also depend on the complexity of the individual solution. If it's too cumbersome (and code-polluting) to support something, we can leave it out.

Yeah, Travis is kind of flaky recently. I just restarted the failed builds. This often already helps.

Since this issue is not urgent, I guess we can continue discussion after your vacation. Have a nice and relaxed time. :relaxed:

pthom:
Hi,

I think I found a satisfying workaround against the differences in the build between 32 and 64 bits. It took me some time to understand this, since I had to reverse engineer the implementation of `fplus::invoke`

The root cause was inside `meta.hpp`, where there was a reimplementation for std::is_function. This implementation seems to fails under windows 32 bits, for whatever reason. 
However, if I deactivate this under MSVC, everything is working fine under MSVC 2017 32 and 64 bits (and only one test fails under MSVC 2015)

So I ended up doing something like this inside `meta.hpp`:
````cpp
// there seems to be a bug in libc++'s std::is_function
// provide our own (cppreference one)
// (the MSVC implementation seems correct)
#ifndef _MSC_VER
#define PROVIDE_IS_FUNCTION_POLYFILL
#endif

#ifndef PROVIDE_IS_FUNCTION_POLYFILL
template<class... Any>
using is_function = std::is_function<Any...>;
#else //PROVIDE_IS_FUNCTION_POLYFILL
// primary template
template<class>
struct is_function : std::false_type { };

// specialization for regular functions
template<class Ret, class... Args>
struct is_function<Ret(Args...)> : std::true_type {};
...
````

So, in my opinion, this PR is perhaps almost ready to merge. 

You will need to create an account at appveyor for the continuous integration under MSVC, and optionaly you can also add a badge for the windows build status on the readme file (see [appveyor status badges](https://www.appveyor.com/docs/status-badges/)

Dobiasd:
Wow, this is extraordinary investigation work! :bowing_man:

Just looked though your changes again, and they seem like strict improvements to me. Really very well done. :+1:

---

I created an appveyor account and added FunctionalPlus: https://ci.appveyor.com/project/Dobiasd/functionalplus

So I guess after merging I can just add the badge like this to the README file: `https://ci.appveyor.com/api/projects/status/github/dobiasd/FunctionalPlus` to 

---

Shall we merge, or is there something left you would like to do before?

pthom:
Thanks for the kind words :-)

I think we can merge ! 
What I have not tested is whether the appveyor build is automatically triggered after each push, and if the github ui will actually display the windows build failures. But I guess this can be tested  a dummy commit where you add something like

````
#ifdef _MSC_VER
#error
#endif
````

to fplus.hpp

Thanks for your support !

Dobiasd:
AppVeyor badge seems to be working:
image

Let's see how the builds turn out. :)

Dobiasd:
:tada: 

Thank you so much again for this splendid contribution. :)

PR #140, opened 2018-10-07
Timed

pthom:
Hi,

Here is a first proposition for an implementation of the profiling decorator.
I added a new type `timed<T>` +  a `make_timed_function<Fn>(Fn f>`decorator. 

This is still work in progress, but your input would be welcomed

Cheers


Dobiasd:
As expected, this looks quite nice. :+1:
I started with reading the tests to get a feel for the API. Seems comfortable to use.

(Btw. thanks for also adding a test for `stopwatch` and the good re-use of existing fplus functions, e.g., `mean_stddev`, in the new tests.)

Regarding the memory management using `std::array<unsigned char, sizeof(T)> mem_;` instead of, let's say, `T val_` or `std::unique_ptr<T>`: In `fplus::maybe` I did this to:
- allow for optionality
- while keeping the value coherent to the wrapping structure in memory for performance. Prefer stack memory over heap allocations there. (Profiling an actual application revealed this bottleneck.)

However in the case of `timed` I'm not sure if this additional code complexity is worth it:
- We don't need optionality, i.e., just use `timed<maybe<T>>` if you do.
- I imagine one usually does not have thousands of timed objects in a performance-critical section, but uses timing for things that actually take time to produce/calculate. And in that case the heap-memory-management-performance hit usually would not be relevant. If true, the code perhaps could be simplified.

What do you think?

pthom:
Hi,

Thanks for your review! 

I modified the timed class in order to store a `T`value. The code is simpler and easier to debug also.

However, there is a corner case in which it does not work, and I would appreciate your input.

Look at the following sample usage:

````cpp
    {
        using Ints = std::vector<int>;

        Ints ascending_numbers = fplus::numbers(0, 10);
        Ints shuffled_numbers = fplus::shuffle(std::mt19937::default_seed, ascending_numbers);

        auto sort_bench = fplus::make_timed_function(fplus::sort<Ints>);
        auto sorted_numbers = sort_bench(shuffled_numbers);
        REQUIRE_EQ(sorted_numbers.get(), ascending_numbers);
    }

````

I get an error at the line `auto sorted_numbers = sort_bench(shuffled_numbers)`
:
````
FunctionalPlus\include\fplus/timed.hpp(74): error C2664: 'std::vector<int,std::allocator<_Kty>> (Container &&)': cannot convert argument 1 from 'Ints' to 'Ints &&'
          with
          [
              _Kty=int,
              Container=Ints
          ]
  FunctionalPlus\include\fplus/timed.hpp(74): note: You cannot bind an lvalue to an rvalue reference
etc... (template noise)
````

This is due to the signature of `fplus::sort`, which avoids unecessary copies:
````cpp
template <typename Container,
    typename ContainerOut = internal::remove_const_and_ref_t<Container>>
ContainerOut sort(Container&& xs)
{
    typedef typename std::remove_reference<Container>::type::value_type T;
    return sort_by(std::less<T>(), std::forward<Container>(xs));
}
````

So, I guess a trick like `internal::remove_const_and_ref_t` might be required; but I do not know what is the best way to do it.

Do you have any idea?

Thanks!

**Comment by Dobiasd** [comment]
The `Container&&` in `ContainerOut sort(Container&& xs)` is not an rvalue reference. It's what Scott Meyers calls a "universal reference". The compiler can deduce it to be an lvalue reference, like in the following case:
```cpp
const std::vector<int> xs = {1, 4, 2, 3};
const auto ys = fplus::sort(xs);
```

Replacing
```cpp
auto sort_bench = fplus::make_timed_function(fplus::sort<Ints>);
```

with
```cpp
auto sort_func = [](const Ints& values){ return fplus::sort(values); };
auto sort_bench = fplus::make_timed_function(sort_func);
```

might help?

pthom:
Hello again,

I started to think about the API for a possible benchmark session. 
My feeling is that it might be better to also provide a way to transform a function into another 
without modifying its signature. 

This way, it would be easier to integrate benchmarking into parts or all of an existing algorithm (otherwise, things could get complex quickly for the user: there is more than just composition in algorithms :-)

For example, with the following `make_bench_function` (work in progress):
````cpp
    template<class Fn>
    auto make_bench_function(benchmark_session & session, const FunctionName & name, Fn f)
    {
        // shall transform f into a function with the same 
        // signature, that will store timings into the benchmark session
        return f;
    }
````

As a toy example, I worked on your Gemstones code, and I imagined that someone would for example need to benchmarks only parts of it (`split_lines` and `convert_container` for example):

Here is the original code:
````cpp
std::string gemstones(const std::string& input)
{
    using namespace fplus;

    typedef std::set<std::string::value_type> character_set;

    // Get lines from input.
    const auto lines = split_lines(false, input);

    // Convert string lines into sets of chars.
    const auto sets = transform( convert_container<character_set, std::string>, lines);

    // Build the intersection of all character sets.
    const auto gem_elements = fold_left_1(
        set_intersection<character_set>, sets);

    // Convert gem_elements.size() into a std::string.
    return show(size_of_cont(gem_elements));
}
````

and here would be the benchmarked code (where `my_benchmark_session` is a variable that will hold the timings)

````cpp
std::string gemstones_bench(const std::string& input)
{
    using namespace fplus;

    typedef std::set<std::string::value_type> character_set;

    // In this example, we only benchmark 2 functions : split_lines and convert_container
    auto split_lines_bench = 
        make_bench_function(my_benchmark_session, "split_lines", split_lines<std::string>);
    auto convert_container_bench = 
        make_bench_function(my_benchmark_session, "convert_charset_string", convert_container<character_set, std::string>);

    // Get lines from input.
    // const auto lines = split_lines(false, input);
    const auto lines = split_lines_bench(false, input);

    // Convert string lines into sets of chars.
    //const auto sets = transform( convert_container<character_set, std::string>, lines);
    const auto sets = transform(convert_container_bench, lines);

    // Build the intersection of all character sets.
    const auto gem_elements = fold_left_1(
        set_intersection<character_set>, sets);

    // Convert gem_elements.size() into a std::string.
    return show(size_of_cont(gem_elements));
}

````

Below, here is my current idea of what benchmark_session should look like:

````cpp

namespace fplus
{
    using FunctionName = std::string;

    // benchmark_session stores timings during a benchmark session
    // and is able to emit a report at the end
    class benchmark_session
    {
    public:
        // report() shall return a string with a summary of the session
        // example below:
        //
        // Times are in ms
        // Function                    |   NbCalls   |   Total Time(ms) |   Av. Time(ns)  |   Deviation(ns)
        // ----------------------------+-------------+------------------+-----------------+--------------------
        // split_lines                 |   200       |   300ms          |   15000ns       |     3000ns
        // convert_charset_string      |   500       |  1500ms          |   3000ns        |     1000ns
        std::string report() const { return "work in  progress"; }

        inline void store_one_time(const FunctionName & function_name, ExecutionTime time) {
            times_[function_name].push_back(time);
        }
    private:
        std::map<FunctionName, std::vector<ExecutionTime>> times_;
    };

    template<class Fn>
    auto make_bench_function(benchmark_session & session, const FunctionName & name, Fn f)
    {
        // shall transform f into a function with the same 
        // signature, that will store timings into the benchmark session
        return f;
    }

    template<class Fn>
    auto run_n_times(int nb_runs, Fn f)
    {
        // run_n_times is a function I would have liked to find in fplus :-)
        for (auto _ : fplus::numbers(0, nb_runs))
            f();
    }

}

````

And it could be used like this

````cpp
// static benchmark session
fplus::benchmark_session my_benchmark_session;

void run_my_benchmark()
{
    std::string input("Lorem ipsum\ndolor sit amet,\nconsectetur,\nadipisci velit");
    fplus::run_n_times(10000, [&]() { gemstones_bench(input); });
    fplus::show(my_benchmark_session.report());
}

````

What do you think? 

PS: 
I still have doubts about whether it is a good idea to develop such a benchmark_session, since may be it goes too deep into the job of a real profiler :-)

Dobiasd:
While the original functor-monad-ish idea for benchmarking was tempting, the simpler approach, with the proposed benchmark object might really be easier to integrate for the end user.

The downside of course is, functions produced by `make_bench_function` have a side effect, that can not be seen from the type signature. However with returning the `timed` object, it would be more obvious.

And yes, with the benchmark session the overlap to profilers seems quite big. In my projects usually when I needed to measure duration of a function call, it was mostly for logging purposes, and I would be happier to have the result in the returned objects (or perhaps `pair<Duration, T>`) instead of having to get it up in some (benchmark) structure.

What use cases do you have in mind?

pthom:
I rewrote the history of the timed branch in order to make its history more clear (sorry).

Here is what I did:

* Add another test around `sort_bench` (your suggestion works perfectly)
* Add some API doc for `make_timed_function` (with an example of usage)
* Make `timed<T>` a public derivate from `std::pair<T, double>` (do you think it is better like this ?)

Concerning the benchmark session I agree with you concerning the overlap with profilers. 
The use case I was thinking about is when you need to profile only parts of a complex algorithm (it is not always easy to start/stop a profiler only in some parts of the code). But, this was just an hypothetical use case.

Anyhow, `make_timed_function` alone might be a sensible addition. 

My idea about a benchmark session is a separate subject I think. I will work some more time on it, and ponder whether it is useful or not.

Dobiasd:
Yes, seems reasonable. We can have `make_timed_function` and still have the option to add benchmark sessions later.

pthom:
So, I tried to conclude on `make_timed_function` with the 2 last commits:

* CMakeLists : add timed_test and stopwatch_test into add_custom_target (https://github.com/Dobiasd/FunctionalPlus/pull/140/commits/672e5e926525cd78d39ec4ccbd92d37ff5cc1bf8)
* fplus.htpp includes timed.hpp (https://github.com/Dobiasd/FunctionalPlus/pull/140/commits/d3066359c175b5fac6f74b0d208eab97f9fad137)

Tell me if anything more is needed before merging.

Thanks!

pthom:
Concerning the benchmark session, I continued to work on it on a separate branch. 

I am still not convinced it is a good idea, but since I was in a good flow, I worked in order to finish a first version. You can see it at https://github.com/pthom/FunctionalPlus/pull/1

It can give the following output:
````
> ./test/benchmark_session_test
Function              |Nb calls|Total time|Av. time|Deviation|
----------------------+--------+----------+--------+---------+
convert_charset_string|    4000|   4.942ms| 1.236ns|  1.390ns|
split_lines           |    1000|   4.528ms| 4.528ns|  1.896ns|
===============================================================================
[doctest] test cases:      1 |      1 passed |      0 failed |      0 skipped
````

From a test file like this: https://github.com/pthom/FunctionalPlus/blob/e33e3b629c31979491987d0aa595313bba2778f9/test/benchmark_session_test.cpp

pthom:
Hum, this PR is not ready yet. I still need to make some adaptations, as `make_timed_function` does not work with functions that return `void`.

Dobiasd:
Ah yeah, the good old problem of C++ not having an actual unit type. : /

pthom:
Ok, if you look at https://github.com/Dobiasd/FunctionalPlus/pull/140/commits/f215fb7f6ed3b180fc4cec8233b028c4407fd0f2 you will see that I added a function `make_timed_void_function`+ its tests.

However, I did it by adding a new unit type `fplus::internal:: UnaryVoid`

Do you think this is overkill ? May be it would be better to let  `make_timed_void_function` transform a void function into a function that returns a double ?

Do you think it would be feasible / sane to detect at compile time whether `make_timed_void_function` or `make_time_function` should be called ? If so, I guess, I could look for a template trick with something like `std::is_same`; but I'm afraid it might get complicated.

Dobiasd:
I think `make_timed_void_function` is OK. Then we don't need `UnaryVoid`. Also I would not provide a compile-time dispatch between `make_timed_function` and `make_timed_void_function`. In C++ one can not store the result of a `void` function, so for me it's OK if the user has to handle this explicitly. One also can not apply `fplus::transform` with a `void` function. So it's consistent in my opinion.

pthom:
Ok, I removed UnaryVoid and made so that `make_timed_void_function(f)` returns a function that returns double.

Do you think it is appropriate that I do a rewrite of the history of this branch in order to make it more clear, i.e for example remove from the history the commit that included UnaryVoid?

Dobiasd:
You can do it if you like, but for me personally this is not mandatory.

Dobiasd:
I can't think of anything right now. I'll review and possibly merge tomorrow, if you agree.

pthom:
of course !

Dobiasd:
Look good. :)

Just one minor thing and one question:
- The `API search type` of `show_timed` has `show_maybe` in it.
- Currently fplus::timed<double>::time() returns the duration in seconds as double. However neither the return type (`double`) nor the function name currently communicate the used unit. How do you feel about using `std::chrono::duration` for this? Otherwise we could maybe just change the name to `fplus::timed<double>::time_in_s()`?

pthom:
> Look good. :)

Thanks!

> The API search type of show_timed has show_maybe in it.

I corrected this, sorry for the typo

>Currently `fplus::timed<double>::time()` returns the duration in seconds as double. 
>However neither the return type (double) nor the function name currently communicate the used unit. > How do you feel about using `std::chrono::duration` for this? Otherwise we could maybe just change > the name to `fplus::timed<double>::time_in_s()`?

I changed the name to `time_in_s()`

Concerning `std::chrono::duration`, my opinion is that the std::chrono API is, well, quite intricate and intimidating for beginners. 

So I also provided `std::chrono::duration<double, std::ratio<1>> duration_in_s()` but it is a supplement to `time_in_s()`. Is this ok for you?

pthom:
Thanks for the merge 👍 
Do you plan update the api site (http://www.editgym.com/fplus-api-search/) so that a search for timed leads to `make_timed_function` ?

PR #141, opened 2018-10-12
format make_timed_function api doc

pthom:
Hello

This is a simple PR in which I try to improve the doc rendering in the editgym website : 

* place code comments at end of code lines (instead of comment on the line before)
* add a blank line at the end of the doc ( with a trailing space after `//`, since I guess that is what is needed in order to hide the `//`)

image

### DIFF
--- a/include/fplus/timed.hpp
+++ b/include/fplus/timed.hpp
@@ -76,9 +76,9 @@ namespace internal
 // API search type: make_timed_function : ((a -> b)) -> (a -> Timed b)
 // fwd bind count: 0
 // Transforms a function into a timed / benchmarked version of the same function.
-// 
+// -
 // Example:
-// 
+// -
 // using Ints = std::vector<int>;
 // Ints ascending_numbers = fplus::numbers(0, 1000);
 // Ints shuffled_numbers = fplus::shuffle(std::mt19937::default_seed, ascending_numbers);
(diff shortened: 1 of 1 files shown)


Dobiasd:
In my opinion adding trailing spaced should not be part of the solution. Many people (including me) have their editors configured in a way to automatically remove trailing spaces for each line. :wink: 

I will experiment a bit, maybe adjusting [`parse_source_files.cpp`](https://github.com/Dobiasd/FunctionalPlus/blob/master/api_search/parse_source_files.cpp).

Dobiasd:
A cheap solution would be to just use `// -`.

image

How do you feel about this?

pthom:
I agree with you & pushed the modif

include/fplus/timed.hpp
```diff
@@ -124,16 +123,15 @@ namespace internal
...
-// void foo() {
-//     std::this_thread::sleep_for(std::chrono::milliseconds(1000));
-// }
+// -
+// void foo() { std::this_thread::sleep_for(std::chrono::milliseconds(1000)); }
 // ...
 // auto foo_bench = make_timed_void_function(foo);
 // auto r = foo_bench();
 // double run_time = foo_bench(); // in seconds
+// -
```
Dobiasd: You can remove this one line.The spacing between the documentation and the function declaration is added automatically by the frontend.

Dobiasd:
So this looks good now. Shall we merge?

pthom:
yes !

Dobiasd:
Looks good. :+1:

image

PR #142, opened 2018-10-12
Benchmark session

pthom:
As discussed, here is my take on benchmark_session.

I took some time to think about it and I found three possible use cases:

- multiplatform development: when you port some code on a different platform (windows, iOs, etc), and you want to quickly profile without having to learn/install/buy a profiler for this platform.

- on site benchmarking : it could be feasible to deploy an application on a customer site (or on a server) where carefuly selected parts of the app are benchmarked. 

- CI, unit-test and integration test: when running longer tests in CI (integration test), it could be desirable to assert that the original performance is kept, and to be able to pinpoint the cause if it is not kept. I discussed this possibility with one of my lead developers, and his answer was 
that he might consider using something like this (especially in CI, when outside of his main platform). He was also veru interested in the `make_timed_function` (last pull request)

A minimal `benchmark_session`could look like:
````cpp
fplus::benchmark_session benchmark_sess;
void foo() {
    auto add_bench = fplus::make_benchmark_function(benchmark_sess, "add", add);
    auto printf_bench = fplus::make_benchmark_void_function(benchmark_sess, "printf", printf);
    int forty_five = add_bench(20, add_bench(19, 6));
    int forty_two = benchmark_expression(benchmark_sess, "sub", forty_five - 3);
    printf_bench("forty_two is %i\n", forty_two);
}
void main() {
    foo();
    std::cout << benchmark_sess.report();
}
````
And it gives the following report:
````
Function|Nb calls|Total time|Av. time|Deviation|
--------+--------+----------+--------+---------+
printf  |       1|   0.010ms| 9.952ns|  0.000ns|
add     |       2|   0.000ms| 0.050ns|  0.009ns|
sub     |       1|   0.000ms| 0.039ns|  0.000ns|
````

As a side note, I added a macro (this is bad, I know, I know...) :

````cpp
#define benchmark_expression(bench_session, name, expression)      \
    make_benchmark_function(                                       \
        bench_session,                                             \
        name,                                                      \
        [&]() { return expression; }                               \
    )();
````

It is suprisingly useful, because you can also benchmark an expression without having to create a lambda/function.
For example, if you want to benchmark the following line:

````cpp
    auto shuffled_numbers = fplus::shuffle(std::mt19937::default_seed, ascending_numbers);
````

You can do it like this

````cpp
    auto shuffled_numbers = benchmark_expression(
        my_benchmark_session,
        "shuffle",
        fplus::shuffle(std::mt19937::default_seed, ascending_numbers); // same line as before
    );
````

I know that this PR is not really in the original scope of fplus, so your opinion will matter to me.

### Remark

benchmark.hpp contains two functions that might be of interest for fplus.
What do you think? If you think they are interesting, I can make a separate pull request for them.

````cpp
    // transform_and_sum is perhaps a nice addition to fplus 
    // (this is equivalent to transform_reduce with the default empty value and + operator)
    template <typename UnaryF, typename Container>
    auto transform_and_sum(UnaryF unary_f, const Container& xs)
    {
        return fplus::sum(fplus::transform(unary_f, xs));
    }

    template<class Fn>
    auto run_n_times(int nb_runs, Fn f)
    {
        for (auto _ : fplus::numbers(0, nb_runs)) {
            (void) _; // suppress warning / unused variable
            f();
        }
    }
````

### DIFF
--- a/include/fplus/benchmark_session.hpp
+++ b/include/fplus/benchmark_session.hpp
@@ -264,13 +264,6 @@ make_benchmark_void_function(                                      \
 
 namespace internal
 {
-    // transform_and_sum is perhaps a nice addition to fplus
-    template <typename UnaryF, typename Container>
-    auto transform_and_sum(UnaryF unary_f, const Container& xs)
-    {
-        return fplus::sum(fplus::transform(unary_f, xs));
-    }
-
     inline std::string show_table(const std::vector<std::vector<std::string>>& rows)
     {
         if (rows.empty() || rows[0].empty())
@@ -299,12 +292,12 @@ namespace internal
         };
 
         auto show_one_row = [&](const std::vector<std::string> & row) {
-            return fplus::internal::transform_and_sum(
+            return fplus::sum(fplus::transform(
                 show_one_element, 
-                fplus::zip(row, columns_width));
+                fplus::zip(row, columns_width)));
         };
 
-        auto firstrow_separator = fplus::internal::transform_and_sum(show_one_separator, columns_width);
+        auto firstrow_separator = fplus::sum(fplus::transform(show_one_separator, columns_width));
         auto rows_formatted = fplus::transform(show_one_row, rows);
         auto rows_separated = fplus::insert_at_idx(1, firstrow_separator, rows_formatted);
         return fplus::join( std::string("\n"), rows_separated) + "\n";
--- a/include/fplus/benchmark_session.hpp
+++ b/include/fplus/benchmark_session.hpp
@@ -114,6 +114,14 @@ namespace fplus
         return bench_function_impl<Fn>(session, name, f);
     }
 
+#define benchmark_expression(bench_session, name, expression)      \
+    make_benchmark_function(                                   \
+    bench_session,                                             \
+    name,                                                      \
+    [&]() { return expression; }                               \
+    )();
+
+#define COMMA ,
 
     template<class Fn>
     auto run_n_times(int nb_runs, Fn f)
--- a/include/fplus/benchmark_session.hpp
+++ b/include/fplus/benchmark_session.hpp
@@ -155,50 +155,48 @@ namespace internal
 // Transforms a function into a function with the *same* signature
 // and behavior, except that it also stores stats into the benchmark session (first parameter),
 // under the name given by the second parameter.
-// (use make_benchmark_void_function if your function returns void).
-// Notes
-// -  make_benchmark_function *will add side effects* to the function
-// (since it stores data into the benchmark session at each call).
-// - If you intend to benchmark only one function, prefer to use the simpler `make_timed_function`
-// --------------------------------------------------
-// Example of a minimal benchmark session:
-// ---------------------------------------
-// fplus::benchmark_session benchmark_sess;
-// void foo() {
-//     auto add_bench = fplus::make_benchmark_function(benchmark_sess, "add", add);
-//     auto printf_bench = fplus::make_benchmark_void_function(benchmark_sess, "printf", printf);
-//     int forty_five = add_bench(20, add_bench(19, 6));
-//     int forty_two = benchmark_expression(benchmark_sess, "sub", forty_five - 3);
-//     printf_bench("forty_two is %i\n", forty_two);
-// }
-// void main() {
-//     foo();
-//     std::cout << benchmark_sess.report();
-// }
-// ----------------------------------------
-// This will output a report like this:
+// -
+// Notes:
+// Side effects: make_benchmark_function *will add side effects* to the function, since it stores data 
+// into the benchmark session at each call.
+// If you intend to benchmark only one function, prefer to use the simpler "make_timed_function"
+// Use "make_benchmark_void_function" if your function returns void
+// -
+// Example of a minimal benchmark session (read benchmark_session_test.cpp for a full example)
+//     fplus::benchmark_session benchmark_sess;
+//     void foo() {
+//         auto add_bench = fplus::make_benchmark_function(benchmark_sess, "add", add);
+//         auto printf_bench = fplus::make_benchmark_void_function(benchmark_sess, "printf", printf);
+//         int forty_five = add_bench(20, add_bench(19, 6));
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: include/fplus/benchmark_session.hpp [hunk 1]
```diff
@@ -0,0 +1,359 @@
...
+#include <vector>
+#include <mutex>
+#include <fplus/fplus.hpp>
+
+namespace fplus
+{
+using FunctionName = std::string;
+struct benchmark_function_report
+{
+    size_t nb_calls;
```
Dobiasd: Please use `std::size_t` instead of `size_t`.

include/fplus/benchmark_session.hpp
```diff
@@ -0,0 +1,359 @@
...
+    ExecutionTime total_time;
+    ExecutionTime average_time;
+    ExecutionTime deviation;
+};
+
+
+namespace internal
+{
+    std::string show_benchmark_function_report(const std::map<FunctionName, benchmark_function_report> & reports);
```
Dobiasd: Feel free to add line breaks between function parameters (if not too ugly) to stay below 80 characters per line.

include/fplus/benchmark_session.hpp
```diff
@@ -0,0 +1,359 @@
...
+        const auto reports = report_list();
+        return fplus::internal::show_benchmark_function_report(reports);
+    }
+
+    std::map<FunctionName, benchmark_function_report> report_list() const
+    {
+        std::lock_guard<std::mutex> lock(functions_times_mutex_);
+        std::map<FunctionName, benchmark_function_report> report;
+        for (const auto & one_function_time : functions_times_)
+            report[one_function_time.first] = make_bench_report(one_function_time.second);
```
Dobiasd: Please use

```cpp
for (...)
{
   loop-body
}
```

Instead of

```cpp
for (...)
   loop-body
```
I know the first version is not strictly/objective better, but matches the code style.

include/fplus/benchmark_session.hpp
```diff
@@ -0,0 +1,359 @@
...
+{
+    // transform_and_sum is perhaps a nice addition to fplus
+    template <typename UnaryF, typename Container>
+    auto transform_and_sum(UnaryF unary_f, const Container& xs)
+    {
+        return fplus::sum(fplus::transform(unary_f, xs));
+    }
+
+
+    inline std::string show_tableau(const std::vector<std::vector<std::string>> & rows)
```
Dobiasd: Bike shedding: Please no space before the `&`. :wink:

Question: How about the term "table" instead of "tableau"?

test/benchmark_session_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+// This is an example on how to use benchmark_session in order to bench separate parts of an algorithm
+
+// We need to instantiate a session into which the stats will be collected
+fplus::benchmark_session my_benchmark_session;
+
+// antic C style qsort (will be benchmarked against std::sort)
+void qsort_vec_int(std::vector<int> & v)
+{
+    auto cmp = [](const void * a, const void * b) {
+    return ( *(static_cast<const int*>(a)) - *(static_cast<const int*>(b)) );
```
Dobiasd: indentation

include/fplus/benchmark_session.hpp
pthom: ok

include/fplus/benchmark_session.hpp
pthom: done

test/benchmark_session_test.cpp
pthom: done

include/fplus/side_effects.hpp
```diff
@@ -269,6 +269,20 @@ auto execute_max_n_times_until_success(std::size_t n,
         replicate(n, effect_to_std_function(eff)));
 }
 
+// API search type: execute_n_times : (Int, Io a)) -> Io Void
```
Dobiasd: I think the other functions use `Io ()` instead of `Io Void`. :)

pthom:
Hi,

Thanks for your review and your comments!

Concerning `execute_n_times()`, I made a single commit (https://github.com/Dobiasd/FunctionalPlus/pull/142/commits/1cc25e4d5271e35bd8509acf17f44f99c6a6ff9e) which can be cherry picked in a different pull request if needed. Please read the API doc I wrote, I hope it is correct.

Concerning `transform_and_sum()`, I understand your point and I replaced its instances by `sum(transform())`

> For the future however we need to be careful not trying to make FunctionalPlus to an "eierlegende Wollmilchsau" (all-in-one device). 😉

... Yes !

include/fplus/side_effects.hpp
pthom: ok, I changed this

pthom:
I also removed the blank lines in the api doc, and I tried to indent the code examples so that they are visually separated from the explanations.

Dobiasd:
Sorry, the thing with indentation and the long `-------` does not work correctly.
image

I know this is a failure on my side of the implementation, but until the parsing (+ maybe frontend) gets an overhaul, we have to keep the comments simpler. :/

I'll be offline(ish) from tomorrow on for about two weeks, but if you like, feel free to improve the situation in [`parse_source_files.cpp`](https://github.com/Dobiasd/FunctionalPlus/blob/master/api_search/parse_source_files.cpp). :grin:
In case you also would like to compile the frontend, you'd need Elm 0.18.0, since I did not yet update the code to work with version 0.19.0.

pthom:
No problem, enjoy your well deserved time out :-)

I'll try to have a look at `parse_source_files`

pthom:
Ok, I simplified the comments in `benchmark_session.hpp`(there was also a typo in the api doc of `execute_n_times`).

I looked at `parse_source_files` and it seems legit. The doc was correctly inserted into `Database.elm`.

My guess is that somehow the elm code will parse the doc with a markdown parser (or something close). So I removed the part that may look like markdown.

And the result seems ok on my side:

image

Dobiasd:
Thanks! This looks good. So I guess we can merge, right?

pthom:
I hope you enjoyed your holiday :-)

Yes, we can merge!

PR #144, opened 2018-10-29
clang 7 & conan 

pthom:
Hi,

Here is simple pull request. It adds two small adaptations :

* compatibility with clang 7 : one new warning had to be disabled. Also clang 7 requires to explicetely link the unit tests with -lstdc++ and -lm.
* can install doctest with conan, when building the unit tests : this is disabled by default; and the conanfile.py default behavior is preserved.


pthom:
I will close this PR, the one by @offa is better for clang 7 support

offa:
It seems reasonable to me to use conan for the installation of doctest. :+1:

pthom:
Yes, I will do a separate PR one yours is merged

PR #146, opened 2018-10-30
conanfile.py : can install doctest from conan

pthom:
Hello,
This PR adds a new cmake option (FPLUS_UNITTEST_USE_CONAN) that enables
to install doctest via conan.

Usage:
> cd build/
> conan install .. -obuild_unittest=True
> cmake .. -DFPLUS_BUILD_UNITTEST=ON -DFPLUS_UNITTEST_USE_CONAN=ON

By default, the conanfile will *not* intall doctest
(an option 'build_unitest' is required).

Note:
I also had to add two libraries for the link (-lstdc++ and -lm)
May be this is only required for osx (but it does not break the linux build)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,12 +5,13 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 project(FunctionalPlus VERSION 0.2.2)
 
 message(STATUS "===( ${PROJECT_NAME} ${PROJECT_VERSION} )===")
-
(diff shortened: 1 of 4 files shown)


offa:
> Note:
I also had to add two libraries for the link (-lstdc++ and -lm)
May be this is only required for osx (but it does not break the linux build)

I'm not sure if this is good. This forces the usages of libstdc++ even one wants to use libc++ (or another std implementation) instead.

Dobiasd:
Oh right. I should have realized this before merging. Thanks offa!

pthom:
Thanks !

Issue #147, opened 2018-10-30
Review the std library problem on OS X


offa:
On Mac OS it's necessary to explicitly add *libstdc++* and *libm* (see #146). This is now [forced](
https://github.com/Dobiasd/FunctionalPlus/blob/e813ab2d791c2cefb057723ed0da9943325b8683/test/CMakeLists.txt#L14) for all Unix like builds.

@pthom Are you building with GCC on Mac OS?

pthom:
I am building with "Apple Clang" (apple fork of clang) and a separate install of clang. 

No gcc though.

I will see if I have time to add an OSX builder to the travis matrix.  It should not be to complicated.

offa:
This is odd, clang should pick up libc++ by default on your system. What error do you get if you build without the workaround (withouth [this line](https://github.com/Dobiasd/FunctionalPlus/blob/master/test/CMakeLists.txt#L14))?

pthom:
Well the problem I encounter seems not to be apple specific.

Sorry if this message is a bit long, but I wanted to make a full analysis.

### Status with "apple clang" : works 

The compilation will work even if I build without the workaround.

````bash
mkdir build-appleclang && cd build-appleclang
conan install .. -obuild_unittest=True && cmake .. -DFPLUS_BUILD_UNITTEST=ON -DFPLUS_UNITTEST_USE_CONAN=ON
make
````

### Status with clang7 under osx 

The compilation will not work without the workaround.

````bash
brew install llvm # will install clang7 into /usr/local/Cellar/llvm/7.0.0/bin
mkdir build-clang7 && cd build-clang7
export CC=/usr/local/Cellar/llvm/7.0.0/bin/clang
export CXX=/usr/local/Cellar/llvm/7.0.0/bin/clang
# export CXXFLAGS="-stdlib=libc++"
conan install .. -obuild_unittest=True && cmake .. -DFPLUS_BUILD_UNITTEST=ON -DFPLUS_UNITTEST_USE_CONAN=ON
make

Undefined symbols for architecture x86_64:
  "std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::compare(unsigned long, unsigned long, char const*, unsigned long) const", referenced from:
      doctest::detail::logTestStart(doctest::detail::TestCase const&) in stringtools_cp1251_test.cpp.o
etc...
````


### Status with clang7 under ubuntu 16.04 / docker

[...]

PR #148, opened 2018-10-30
Enable Std lib workaround for apple only

offa:
This limits the standard lib workaround (#147) to apple builds (mac os etc.) only.


pthom:
This patch works perfectly on my mac 👍

PR #149, opened 2018-10-31
make concat work with generic containers

Pcres91:
Explicitly declare using std::begin and std::end to stop any ambiguous function calls


Dobiasd:
Thanks! :+1:

PR #151, opened 2018-11-08
Added osx travis build

pthom:
cf https://github.com/Dobiasd/FunctionalPlus/issues/147

Hi,

This PR adds two new builders to the travis matrix (xcode 9 and xcode 10)

I had to fight hard against travis inconsistencies, since python support is different on osx.

Details :
* locale tests adapted to osx (different locale names) 
* I had to reduce the timed_test.cpp expectations under osx, since these builders are very slow with travis
* I modified the badges on the readme file, but this commit can be reversed if desired

Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
-language: python
-python: 3.6
-
+language: generic
 dist: trusty
 sudo: required
 
 matrix:
     include:
     - os: linux
+      language: python
+      python: 3.6
       compiler: gcc
       env: GCC_VERSION=8
         - CC=gcc-8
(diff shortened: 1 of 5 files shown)


pthom:
Modif to the readme badges :

Dobiasd:
Nice! But I have to admit, I might not be the best person to judge CI improvements like this. So if @offa does not have any objections, I'll merge. ;)

offa:
Looks good :+1: 

Just an aesthetic note: I don't think the build bagdes need an descriptive text, they are well known on GH :smile: ) (and the readme breaks a bit unfortunate on my phone :confused:  ). However, that doesn't really matter …

We probably should add OS X Clang to the list of [supported compiler](https://github.com/Dobiasd/FunctionalPlus#requirements-and-installation).

offa:
@pthom do you think we can close #147 once this PR has been merged?

Issue #162, opened 2020-02-13
Adding to CMakeLists.txt as a dependency


xmfcx:
Hello,
By following this document:
https://github.com/Dobiasd/FunctionalPlus/blob/master/INSTALL.md
I was able to include it to a project using cmake's ExternalProject method and it works. But I would like to be able to build it without my cmake cloning something on other computers.

I have also followed the first way, cloned, compiled and performed `sudo make install`

How do I include it to my cmakelists file now? Is there a simple way like:
```cmake
find_package(functional_plus)
include_directories(${FPLUS_INCLUDE_DIR})
add_dependencies(main functional_plus)
```

Thanks.

`sudo make install` output:
```bash
Install the project...
-- Install configuration: ""
-- Installing: /usr/local/include/fplus
-- Installing: /usr/local/include/fplus/benchmark_session.hpp
-- Installing: /usr/local/include/fplus/generate.hpp
-- Installing: /usr/local/include/fplus/side_effects.hpp
-- Installing: /usr/local/include/fplus/queue.hpp
-- Installing: /usr/local/include/fplus/read.hpp
-- Installing: /usr/local/include/fplus/raii.hpp
-- Installing: /usr/local/include/fplus/container_properties.hpp
-- Installing: /usr/local/include/fplus/composition.hpp
-- Installing: /usr/local/include/fplus/interpolate.hpp
-- Installing: /usr/local/include/fplus/maps.hpp
-- Installing: /usr/local/include/fplus/tree.hpp
-- Installing: /usr/local/include/fplus/stopwatch.hpp
-- Installing: /usr/local/include/fplus/filter.hpp
-- Installing: /usr/local/include/fplus/timed.hpp
-- Installing: /usr/local/include/fplus/maybe.hpp
-- Installing: /usr/local/include/fplus/optimize.hpp
-- Installing: /usr/local/include/fplus/curry.hpp
-- Installing: /usr/local/include/fplus/container_common.hpp
-- Installing: /usr/local/include/fplus/show.hpp
-- Installing: /usr/local/include/fplus/internal
-- Installing: /usr/local/include/fplus/internal/meta.hpp
-- Installing: /usr/local/include/fplus/internal/invoke.hpp
-- Installing: /usr/local/include/fplus/internal/apply.hpp
-- Installing: /usr/local/include/fplus/internal/composition.hpp
-- Installing: /usr/local/include/fplus/internal/container_common.hpp
[...]

PR #163, opened 2020-02-15
Add an additional instruction how to use FunctionalPlus after installing.

kkweon:
Adding instruction on how to add **FunctionalPlus** into other cmake projects after the manual installation

## Related

#162


Dobiasd:
Perfect! Thank you very much. :heart:

PR #168, opened 2020-04-01
Doctest v2.3.4

offa:
Doctest v2.2.3 wasn't found by conan (it's not available any more!?)

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -15,7 +15,7 @@ class FunctionalPlusConan(ConanFile):
 
     def requirements(self):
         if self.options.build_unittest:
-            self.requires.add('doctest/2.2.3@bincrafters/stable')
+            self.requires.add('doctest/2.3.4@bincrafters/stable')
 
     def package(self):
         self.copy("*LICENSE", dst="licenses")


Dobiasd:
I was expecting old versions to usually still be available, but ok. :man_shrugging:

Thanks a lot for the fix. :heart:

offa:
I wonder why travis doesn't show up!? :question:

Dobiasd:
Looks OK now:

gND6qJZ

Nybx4vt

offa:
But the status isn't shown on Github (only appveyor).

Dobiasd:
I think recently, it only shows the green or the red status, but no longer the yellow one.

We could make `master` fail to check if it updates then. :wink:

Dobiasd:
Or, maybe by some internal magic it remembers that this build already passed when it ran on the branch, so it does not change the status while (re-)building. Also only testable with pushing some dummy commit to `master` directly, I guess.

offa:
Seems others experience the problem too: https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798

Dobiasd:
Ah, so maybe we should [migrate](https://docs.travis-ci.com/user/migrate/legacy-services-to-github-apps-migration-guide/), I guess.

offa:
No, legacy build still work and should post their result :-)
I hope travis devs fix this soon. But until then I guess we can't do much.

Issue #170, opened 2020-04-05
Error when compiling with GCC 10


ruanformigoni:
Greetings,

When using GCC trunk, the code in `split.hpp` does not compile, here is the error output:

<details><summary>Error Output</summary>
/home/ruan/.conan/data/functionalplus/v0.2.6-p0/dobiasd/stable/package/5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9/include/fplus/split.hpp: In instantiation of ‘ContainerOut fplus::split_by(UnaryPredicate, bool, const ContainerIn&) [with UnaryPredicate = bool (*)(const char&); ContainerIn = std::__cxx11::basic_string<char>; ContainerOut = std::vector<std::__cxx11::basic_string<char> >]’:
/home/ruan/.conan/data/functionalplus/v0.2.6-p0/dobiasd/stable/package/5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9/include/fplus/string_tools.hpp:76:20:   required from ‘ContainerOut fplus::split_lines(bool, const String&) [with String = std::__cxx11::basic_string<char>; ContainerOut = std::vector<std::__cxx11::basic_string<char> >]’
/home/ruan/.conan/data/functionalplus/v0.2.6-p0/dobiasd/stable/package/5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9/include/fplus/side_effects.hpp:428:76:   required from here
/home/ruan/.conan/data/functionalplus/v0.2.6-p0/dobiasd/stable/package/5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9/include/fplus/split.hpp:342:20: error: ambiguous overload for ‘operator=’ (operand types are ‘std::insert_iterator<std::vector<std::__cxx11::basic_string<char> > >’ and ‘<brace-enclosed initializer list>’)
  342 |             *itOut = {};
      |             ~~~~~~~^~~~
In file included from /home/ruan/GCC/include/c++/10.0.1/iterator:63,
                 from /home/ruan/.conan/data/fmt/6.1.2/_/_/package/b173bbda18164d49a449ffadc1c9e817f49e819d/include/fmt/core.h:13,
                 from /home/ruan/.conan/data/fmt/6.1.2/_/_/package/b173bbda18164d49a449ffadc1c9e817f49e819d/include/fmt/format.h:36,
[...]

PR #171, opened 2020-04-05
Fixed split_by compilation error for GCC 10

ruanformigoni:
Pull request related to issue #170 .

### DIFF
--- a/include/fplus/split.hpp
+++ b/include/fplus/split.hpp
@@ -339,7 +339,7 @@ ContainerOut split_by
         start = internal::add_to_iterator(stop);
         if (allow_empty && start == std::end(xs))
         {
-            *itOut = {};
+            *itOut = std::insert_iterator<ContainerOut>{};
         }
     }
     return result;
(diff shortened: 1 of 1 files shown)


Dobiasd:
Hi Ruan,

thanks a lot for submitting this fix.

However, the Microsoft compiler (MSVC) [seems to not like](https://ci.appveyor.com/project/Dobiasd/functionalplus/builds/31962418/job/nvrtv43em8et1eif) that version using `*itOut = std::insert_iterator<ContainerOut>{};`. And idea how we could make it work for that too without introducing some `#ifdef _MSC_VER` stuff?

Dobiasd:
Ah, sorry, it's not just the MSVC build that fails, but GCC and Clang fail too:

https://travis-ci.org/github/Dobiasd/FunctionalPlus/builds/671193078

Dobiasd:
How about `*itOut = typename ContainerOut::value_type();` instead?

ruanformigoni:
Nice one, it worked! Thank you! :grin:

Dobiasd:
Cool. Squashed and merged. Thanks again for bringing this up. :+1:

PR #172, opened 2020-04-24
GCC 10 ci build

offa:
It's not officially released yet, but already available on an ever increasing number of distributions.


Dobiasd:
:heart:

PR #174, opened 2020-07-01
feat: Nicer usage as CMake subdirectory

henryiii:
It's nice to provide aliases matching exported target names. Would make https://github.com/Dobiasd/frugally-deep/pull/232 simpler.

For example, see https://gitlab.com/libeigen/eigen/-/blob/9cb8771e9c4a1f44ba59741c9fac495d1872bb25/CMakeLists.txt#L582


Dobiasd:
Good idea, thanks!

Issue #175, opened 2020-07-07
In conan-center, provide CMake targets if possible


ruanformigoni:
Greetings,

This issue is a CMake request to enable `find_package`, for the conan-center releases. Currently, this is not possible:

```cmake

if(NOT FunctionalPlus_FOUND)
  find_package(FunctionalPlus REQUIRED)
endif()

target_link_libraries(mylib FunctionalPlus::fplus Catch2::Catch2)
```

I've change conanfile.py to adhere these requests, in my [bintray](https://bintray.com/beta/#/ruanformigoni/functionalplus/functionalplus:ruanformigoni?tab=overview), and it worked with some simple changes. Furthermore, created a pull request #176.

Dobiasd:
Fixed with [pull/176](https://github.com/Dobiasd/FunctionalPlus/pull/176).

PR #176, opened 2020-07-07
Provide CMake targets for conan users

ruanformigoni:
Greetings,

This pull request is for issue #175.


conanfile.py
```diff
@@ -17,8 +17,13 @@ def requirements(self):
...
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.install()
+
     def package(self):
         self.copy("*LICENSE", dst="licenses")
-        self.copy("*.h", src=".")
-        self.copy("*.hpp", src=".")
+        self.copy("*.hpp", src="include")
+        self.copy("*.hpp", src="lib")
```
ruanformigoni, line 28: This includes the files:
* lib/cmake/FunctionalPlus/FunctionalPlusConfig.cmake
* lib/cmake/FunctionalPlus/FunctionalPlusConfigVersion.cmake
* lib/cmake/FunctionalPlus/FunctionalPlusTargets.cmake

conanfile.py
```diff
@@ -1,12 +1,12 @@
...
+from conans import ConanFile, CMake
 
 class FunctionalPlusConan(ConanFile):
     name = "functionalplus"
+    version = "v0.2.8-p0"
     license = "Boost Software License 1.0"
     url = "https://github.com/Dobiasd/FunctionalPlus"
     description = "Functional Programming Library for C++. Write concise and readable C++ code."
-    exports_sources = ["include*", "LICENSE"]
+    exports_sources = ["*", "LICENSE"]
```
ruanformigoni, line 9: The project needs to be installed for the `cmake` files to be generated, therefore, all files are to the `conan` build directory. Although, not all may be necessary.

conanfile.py
```diff
@@ -17,8 +17,13 @@ def requirements(self):
         if self.options.build_unittest:
             self.requires.add('doctest/2.3.4@bincrafters/stable')
 
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.install()
```
ruanformigoni, line 23: Comands of conan to install the `cmake` targets, no need for the build phase.

Dobiasd:
Thanks!
Any idea why the GCC-7 build fails on Travis?

ruanformigoni:
> Thanks!
> Any idea why the GCC-7 build fails on Travis?

I'm looking into it, it seems to be a problem with version mismatching.

conan_build.py
```diff
@@ -5,7 +5,7 @@
     if os.getenv("CXX") == "g++-7":
         version = os.getenv("TRAVIS_TAG")
         if not version:
-            version = "dev"
+            version = "v0.2.8-p0"
```
ruanformigoni, line 8: It was just a version mismatch.

Dobiasd:
@pthom You know more about the build environment than I do. Would you like to take a look at this PR and let us know what you think?

Dobiasd:
@danimtb Could you have a look at this PR and let us know what you think? :slightly_smiling_face:

Dobiasd:
@ruanformigoni Sorry for the delay. As you might have guessed, my knowledge about this is not yet good enough to decide. So maybe you can help me understand things better. I've added some review remarks to the commits, and also have one additional question: Do you think similar changed would also make sense in the [`frugally-deep` repo](https://github.com/Dobiasd/frugally-deep)?

ruanformigoni:
> @ruanformigoni Sorry for the delay. As you might have guessed, my knowledge about this is not yet good enough to decide. So maybe you can help me understand things better. I've added some review remarks to the commits, and also have one additional question: Do you think similar changed would also make sense in the [`frugally-deep` repo](https://github.com/Dobiasd/frugally-deep)?

Hi @Dobiasd , no problem! I didn't know about `frugally-deep`, now I'm sure to use it in my work :smile: . How can I view your remarks? I'm a bit puzzled about the `conan_build.py` file, as it seems to apply to only `g++-7`. Was there a problem with this version of gcc that needed to be manually addressed? I think it would be applicable, this way the user gains more control over the compile flags applied to each library, with `cmake`. Since I opened this issue, I discovered a new feature of conan called [components](https://docs.conan.io/en/latest/reference/generators/cmake_find_package.html?highlight=components), It is a new conan feature which introduces the idom `<PKG-NAME>::<COMP-NAME>`. This way users don't have to look into the docs for the alias and target name, since it is the name of the package and the component, which yields: `functionalplus::functionalplus`, this is the same name the user sees in the `conan search` command. Are you interested in this feature? I could change my pull request to adhere to this.

Dobiasd:
> How can I view your remarks?

Lol, I wrote the remarks, but forgot to click the "submit review" button. -_- Now it's submitted.

> I'm a bit puzzled about the `conan_build.py` file, as it seems to apply to only `g++-7`. Was there a problem with this version of gcc that needed to be manually addressed?

I'm not 100% sure about it, but I guess it was [added that way](https://github.com/Dobiasd/FunctionalPlus/commit/100c9d7ac7e6f73fe7a209d1a7d0c1f67627b1fc#diff-c2944b988c041e4c495df918bf61e041) to built the release package for Conan only once in the CI runs on Travis (which runs for multiple compiler versions). So in that case it would not be something specific to `g++-7`, just some kind of hack to archive exactly-only semantics.

> Since I opened this issue, I discovered a new feature of Conan called components [...] Are you interested in this feature? I could change my pull request to adhere to this.

Sounds good. If it's not too much work for you and does not increase the complexity a lot, it would be cool if you could add it. Thanks. :)

conan_build.py
Dobiasd: It has become `v0.2.9-p0` in the meantime. :slightly_smiling_face:

pthom:
> @pthom You know more about the build environment than I do. Would you like to take a look at this PR and let us know what you think?
Sorry for not answering sooner, I did not see you ping before. However, my knowledge about conan is a bit limited, sorry :-)

conanfile.py
```diff
@@ -1,12 +1,12 @@
-from conans import ConanFile
-
+from conans import ConanFile, CMake
```
yezz123, line 1: Can that be generated or add a library or form can give the file JSON Extension ?

conan_build.py
danimtb: no need to update the version here as this is managed by the CI, see *conan_build.py* in this repo

conanfile.py
danimtb: I don't understand this question. What do you mean with JSON extension?

conanfile.py
danimtb: I really don't get this change

conanfile.py
danimtb: Better move this to the `def pacakge()` command

danimtb:
Sorry for the delay and my comments up there, I didn't check the updated comments in this conversation.

Totally agree with the latest comments here https://github.com/Dobiasd/FunctionalPlus/pull/176#issuecomment-670212293. There is no need to copy the CMake files in the package as Conan already provides a way to model the library targets through components. This is not a substitute for CMake files, but a general model of internal libraries of the package. That way Conan is able to generate files in a suitable format by consumers' demand.

For example, translating that to pkg-config files or visual studio props.

@ruanformigoni the syntax of components is not complex, si I think this PR will benefit from an update on that approach. WDYT? 😄

conanfile.py
ruanformigoni: Hi @Dobiasd, it is the cmake files, but as pointed out by @danimtb, there is no need for it. So I have removed it in the lastest commit.

conanfile.py
ruanformigoni: Hi @Dobiasd  and @danimtb , I have modified `export_sources` to:
```cpp
exports_sources = ["examples*", "cmake*", "include*", "CMakeLists.txt", "LICENSE"]
```
If  I leave just `exports_sources = ["include*", "LICENSE"]`, the compilation fails.

conanfile.py
ruanformigoni: Hi @danimtb, commands moved to `def package()`.

danimtb (approved):

Dobiasd:
With @danimtb (thanks a lot for the review :heart:) having approved this pull request, I guess we might be ready to merge.

@ruanformigoni Do you agree?

ruanformigoni:
> With @danimtb (thanks a lot for the review heart) having approved this pull request, I guess we might be ready to merge.
> 
> @ruanformigoni Do you agree?

I agree :+1:

PR #181, opened 2020-08-09
Use struct instead of using for /permissive-

### LINKED ISSUE #173: Compilation error in result.hpp: 'FOut': is not a class or namespace name
After watching the Udemy course, I tried to integrate fplus in my projects, but I've encountered some errors, which I also encountered in an empty command-line project:
1>C:\Users\...\include\fplus\result.hpp(316,1): error C2653: 'FOut': is not a class or namespace name
1>C:\Users\...\include\fplus\result.hpp(323,1): error C2653: 'FOut': is not a class or namespace name

The code I tried is very simple:

`#include <iostream>`
`#include <fplus/fplus.hpp>`

`int main()`
`{`
`    std::cout << "Hello World!\n";`
`}`

My setup is Visual Studio version 16.5.3, compiler version set to ISO C++ 17 Standard

am11:
Fixes #173.

### DIFF
--- a/include/fplus/result.hpp
+++ b/include/fplus/result.hpp
@@ -304,14 +304,24 @@ auto compose_result(Callables&&... callables)
             internal::trigger_static_asserts<internal::check_arity_tag,
                                                  decltype(f),
                                                  decltype(args)...>();
+#if _MSC_VER >= 1920 // in VS2019, compilation with /permissive- breaks with 'using' syntax below
+            struct FOut : std::decay_t<
+                internal::invoke_result_t<decltype(f), decltype(args)...>> {};
+#else
             using FOut = std::decay_t<
                 internal::invoke_result_t<decltype(f), decltype(args)...>>;
+#endif
 
             internal::trigger_static_asserts<internal::unary_function_tag,
                                                  decltype(g),
                                                  typename FOut::ok_t>();
+#if _MSC_VER >= 1920 // in VS2019, compilation with /permissive- breaks with 'using' syntax below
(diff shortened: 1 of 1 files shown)


Dobiasd:
[On AppVeyor, Visual Studio 2015 is somewhat petulant from that change.](https://ci.appveyor.com/project/Dobiasd/functionalplus/builds/34561530/job/4y2uja5hsrlr4aw1)

```
C:\projects\fplus\include\fplus/result.hpp(308): fatal error C1001: An internal error has occurred in the compiler. [c:\projects\fplus\build\test\result_test.vcxproj]
  (compiler file 'msc1.cpp', line 1468)
   To work around this problem, try simplifying or changing the program near the locations listed above.
  Please choose the Technical Support command on the Visual C++ 
   Help menu, or open the Technical Support help file for more information
  Internal Compiler Error in C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\x86_amd64\CL.exe.  You will be prompted to send an error report to Microsoft later.
```

Do you have an idea how to make it work for that MSVC version too? If not, I suggest we simply drop support for this version.

am11:
Fixed CI breakage by modifying the patch to only use `struct` for VS2019 and above.

Dobiasd:
It's always nicer without `#ifdef`, [but it's not the first time we do this](https://github.com/Dobiasd/FunctionalPlus/search?q=_MSC_VER). :grin:

Thanks a lot for your contribution. I think this will help a lot of people to have a smoother experience with the library. :+1:

PR #189, opened 2020-10-27
add include_all_in_one

pthom:
Hello again,

This PR adds an amalgamated include; so that fplus could also be used on online compilers such as Compiler Explorer.
See for example: https://godbolt.org/z/Tzv377

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,8 @@ target_include_directories(fplus INTERFACE
 find_package(Threads REQUIRED)
 target_link_libraries(fplus INTERFACE Threads::Threads)
 
+add_subdirectory(include_all_in_one)
+
 if(FPLUS_BUILD_UNITTEST)
     enable_testing()
     add_subdirectory(test)
--- a/include_all_in_one/CMakeLists.txt
+++ b/include_all_in_one/CMakeLists.txt
@@ -1,0 +1,4 @@
+execute_process(
+    COMMAND python ${CMAKE_CURRENT_LIST_DIR}/make_all_in_one.py
+    RESULT_VARIABLE out_command
+)
--- a/include_all_in_one/CMakeLists.txt
+++ b/include_all_in_one/CMakeLists.txt
@@ -1,4 +1,5 @@
-execute_process(
+add_custom_target(
+    include_all_in_one
+    ALL
     COMMAND python ${CMAKE_CURRENT_LIST_DIR}/make_all_in_one.py
-    RESULT_VARIABLE out_command
 )
--- a/include_all_in_one/CMakeLists.txt
+++ b/include_all_in_one/CMakeLists.txt
@@ -1,5 +1,10 @@
-add_custom_target(
-    include_all_in_one
-    ALL
-    COMMAND python ${CMAKE_CURRENT_LIST_DIR}/make_all_in_one.py
-)
+find_package (Python COMPONENTS Interpreter QUIET)
+if (Python_FOUND)
+    add_custom_target(
+        include_all_in_one
+        ALL
+        COMMAND ${Python_EXECUTABLE} ${CMAKE_CURRENT_LIST_DIR}/make_all_in_one.py
+    )
+else()
+    message(STATUS "Not generating include_all_in_one, python was not found.")
+endif()
--- a/include_all_in_one/include/fplus/fplus.hpp
+++ b/include_all_in_one/include/fplus/fplus.hpp
@@ -1,0 +1,15930 @@
+
+//
+// fplus.hpp
+//
+
+// Copyright 2015, Tobias Hermann and the FunctionalPlus contributors.
+// https://github.com/Dobiasd/FunctionalPlus
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+
+
+
+//
+// compare.hpp
+//
+
+// Copyright 2015, Tobias Hermann and the FunctionalPlus contributors.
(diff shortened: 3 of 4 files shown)


Dobiasd:
Ah, that's a very good idea. I remember, at one point I was using compiler explorer and had to do some manual copy-pasting to get things to work.

Your solution looks good to me. But I'm wondering if there are any existing tools that already can do this. (`g++ -E` outputs too much, i.e., the full preprocessor output, including the C++ standard headers.) In case there is such a thing, it might be better to use this instead of maintaining our own tool. Did you already search for something like that?

pthom:
No, I did not search at the time I wrote the tool.
I did [a quick google search](https://www.google.com/search?sxsrf=ALeKk01ybDDF_dXt4A3kYqDl5u0kw0z8xQ%3A1603808786091&source=hp&ei=Ei6YX4-eA6WUlwTf-qj4CA&q=c%2B%2B+header+only+amalgamate&oq=c%2B%2B+header+only+amalgamate&gs_lcp=CgZwc3ktYWIQAzIFCCEQoAFQmQRYmQRggwdoAHAAeACAAWSIAWSSAQMwLjGYAQCgAQKgAQGqAQdnd3Mtd2l6&sclient=psy-ab&ved=0ahUKEwiP7-un_dTsAhUlyoUKHV89Co8Q4dUDCAY&uact=5), and I found some alternative existing tools, however I find them a bit cumbersome.

- https://github.com/vinniefalco/Amalgamate : C++, 271 stars, last commit was 9 years ago. Several thousands lines of code (!)
- https://github.com/edlund/amalgamate: python, 52 stars, last comit 6 years ago. Seems pretty easy, but the json config (see [the example json files here](https://github.com/edlund/amalgamate/tree/master/test) might as well be as long as my python source code

Antoher suggestion would be to replace the python script by a short C++ demo program that would showcase fplus abilities.

Dobiasd:
Thanks for the investigation. Yeah, the tools don't look too amazing.

---

> Antoher suggestion would be to replace the python script by a short C++ demo program that would showcase fplus abilities.

You mean using fplus to generate the amalgamated fplus header? Sound fun, but in the end, we would add a similar maintenance cost to the repository. But [the `generate_derived_functions.sh`](https://github.com/Dobiasd/FunctionalPlus/blob/master/generate/generate_derived_functions.sh) is worse. :grimacing:

Maybe I need to better understand the process. Would the amalgamated header be auto-generated when the user does the normal `make install`, or do we (library devs) need to call the script after having changed something in the normal headers and commit-push the resulting amalgamated header to the repo?

pthom:
> Maybe I need to better understand the process. Would the amalgamated header be auto-generated when the user does the normal make install, or do we (library devs) need to call the script after having changed something in the normal headers and commit-push the resulting amalgamated header to the repo?

I just pushed some changes: `include_all_in_one/include/fplus/fplus.hpp` is generated each time you build fplus (previously it was generated only when running cmake). Is that ok for you?

Dobiasd:
That's good too, or maybe even better, thanks.

But why do we have the generated header in the repo then? I'd expect to *either* have it auto-build by cmake/make *or* having it in the repo. Having both means (for the user) it is checked out when doing `git clone` and then overwritten from the (c)make invocation, right?

pthom:
> But why do we have the generated header in the repo then? I'd expect to either have it auto-build by cmake/make or having it in the repo. 

> Having both means (for the user) it is checked out when doing git clone and then overwritten from the (c)make invocation, right?

Yes, you are right, but since the tool generate the header from the same files, the output should be exactly the same as the committed version; especially since the header is generated at each build (so that library developpers do not forget to commit it). 

I just checked that the script works under windows, osx, with either python2 or python3: the generated file is strictly identical to the committed version (I just pushed a modif for python 3, that forces utf-8 encoding).

I do not know if I have the definitive good answer. So let's decide on what we want:

Do we want the generated header to be accessible via an url, for use inside compiler explorer for example? If yes, then I think it should be in the repo, unless we find an alternate idea in order to have it online. If it is not that important, then a call to the script during make install step could be sufficient.

Dobiasd:
> since the header is generated at each build (so that library developpers do not forget to commit it).

Good point. :slightly_smiling_face:

> Do we want the generated header to be accessible via an url, for use inside compiler explorer for example? If yes, then I think it should be in the repo, unless we find an alternate idea in order to have it online. If it is not that important, then a call to the script during make install step could be sufficient.

I guess since using it in compiler explorer is the initial motivation, it makes sense to have it available via a URL. And having it in the repo sounds like to most simple solution, so I'm for it.

---

Ok, thanks for the explanations. I think I just needed a bit more context.

I now think your solution is fine. Let me know when you think we're ready to merge.

pthom:
So, I guess the last question is : do you agree with me adding `add_subdirectory(include_all_in_one)
` in the main CMakeLists.txt with no related global option (e.g. FPLUS_GENERATE_ALL_IN_ONE)?

The reasonning for adding it without any global option is that this step is extremely fast. 

````bash
time python make_all_in_one.py
-- Generating include_all_in_one
python make_all_in_one.py  0.02s user 0.03s system 56% cpu 0.080 total
````

However, it requires the user to have python (2 or 3) in his path.

If you think this is reasonable, then let's merge :-)

Dobiasd:
Mhh, I think needing Python to run `cmake .. && make && sudo make install` for a C++ library might not be ideal.

Would it be possible to skip this step (without an error, just an info log) in case Python is not found? Otherwise, a flag like the suggested `FPLUS_GENERATE_ALL_IN_ONE` would be good, I think.

pthom:
You are right. I pushed a [modif for this](https://github.com/Dobiasd/FunctionalPlus/pull/189/commits/11693230e4e4667762ff80a8c46d1b3f4e7435c8).

PR #192, opened 2020-10-28
find_all_idxs_by : quasi-verbatim overload for vector of bools (fix #190)

### LINKED ISSUE #190: clang osx warning in `find_all_idxs_by`(search.hpp)
(Sorry to drown you with messages today :-)

With apple-clang under osx I have a warning inside search.hpp (which is considered as an error during the build).

````
clang --version
Apple clang version 12.0.0 (clang-1200.0.32.21)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
````

The file `transform_test.cpp` will fail to compile at this line:
````cpp
TEST_CASE("transform_test, keep_if_parallelly")
{
    const std::vector<int> v = {1, 2, 3, 2, 4, 5};
-->    auto result = fplus::keep_if_parallelly(is_even, v);
    REQUIRE_EQ(result, std::vector<int>({2, 2, 4}));
}
````

And the reported warning is inside `find_all_idxs_by` at the line `for (const auto& x : xs)`:

````cpp
template <typename ContainerOut = std::vector<std::size_t>,
        typename UnaryPredicate, typename Container>
ContainerOut find_all_idxs_by(UnaryPredicate p, const Container& xs)
{
    internal::check_unary_predicate_for_container<UnaryPredicate, Container>();
    std::size_t idx = 0;
    ContainerOut result;
    auto itOut = internal::get_back_inserter(result);
    --> for (const auto& x : xs)
    {
        if (internal::invoke(p, x))
            *itOut = idx;
        ++idx;
    }
    return result;
}

````

Here is the complete compilation log:
````cpp
CMakeFiles/transform_test.dir/transform_test.cpp.o -c ../test/transform_test.cpp
In file included from ../test/transform_test.cpp:9:
In file included from ../include/fplus/fplus.hpp:12:
In file included from ../include/fplus/container_properties.hpp:14:
../include/fplus/search.hpp:117:22: error: loop variable 'x' is always a copy because the range of type 'const std::__1::vector<bool, std::__1::allocator<bool> >' does not return a reference [-Werror,-Wrange-loop-analysis]
    for (const auto& x : xs)
                     ^
../include/fplus/transform.hpp:403:23: note: in instantiation of function template specialization 'fplus::find_all_idxs_by<std::__1::vector<unsigned long, std::__1::allocator<unsigned long> >, bool (*)(const bool &), std::__1::vector<bool, std::__1::allocator<bool> > >' requested here
    const auto idxs = find_all_idxs_by(
                      ^
../test/transform_test.cpp:168:26: note: in instantiation of function template specialization 'fplus::keep_if_parallelly<bool (*)(int), std::__1::vector<int, std::__1::allocator<int> > >' requested here
    auto result = fplus::keep_if_parallelly(is_even, v);
                         ^
../include/fplus/search.hpp:117:10: note: use non-reference type 'std::__1::__bit_const_reference<std::__1::vector<bool, std::__1::allocator<bool> > >'
    for (const auto& x : xs)
         ^~~~~~~~~~~~~~~
1 error generated.
[23/82] Building CXX object test/CMakeFiles/container_common_test.dir/container_common_test.cpp.o
ninja: build stopped: subcommand failed.
````

I could silence i with a simple
````cpp
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wrange-loop-analysis"
````

but may be you have a better idea (but I'm afraid it might require some more SFINAE noise...)

pthom:
Except that one cannot take a reference on vector<bool> items

### DIFF
--- a/include/fplus/search.hpp
+++ b/include/fplus/search.hpp
@@ -123,6 +123,25 @@ ContainerOut find_all_idxs_by(UnaryPredicate p, const Container& xs)
     return result;
 }
 
+// find_all_idxs_by : quasi-verbatim overload for vector of bools
+template <typename ContainerOut = std::vector<std::size_t>,
+        typename UnaryPredicate>
+ContainerOut find_all_idxs_by(UnaryPredicate p, const std::vector<bool>& xs)
+{
+    internal::check_unary_predicate_for_container<UnaryPredicate, std::vector<bool>>();
+    std::size_t idx = 0;
+    ContainerOut result;
+    auto itOut = internal::get_back_inserter(result);
+    for (const auto x : xs) // this line differs (cannot take a reference on vector<bool> item)
+    {
+        if (internal::invoke(p, x))
+            *itOut = idx;
+        ++idx;
+    }
+    return result;
+}
+
+
 // API search type: find_all_idxs_of : (a, [a]) -> [Int]
 // fwd bind count: 1
 // Returns the indices off all elements equal to x.


Dobiasd:
This overload will fix `find_all_idxs_by` for `vector<bool>`, but the `for (const auto& x : xs)` thing happens in many other functions, taking a container, too, e.g., `unweave`, `is_subsequence_of`, `keep_idxs`. I guess it does not make much sense to provide (and maintain) a `vector<bool>` overload for all of them.

But simply switching all of them to `for (const auto x : xs)` might also be bad, because it would change the performance characteristics of FunctionalPlus, i.e., some applications that user containers with big/expensive types would suddenly slow down because of all the object copies happening.

How about we instead adjust `keep_if_parallelly` to fix the Clang warning in `TEST_CASE("transform_test, keep_if_parallelly")` (issue #190) by no longer letting it use a temporary `vector<bool>` like [so](https://github.com/Dobiasd/FunctionalPlus/pull/193)?

pthom:
Your proposed fix works and is better. I close this PR.

PR #193, opened 2020-10-28
Avoid a temporary std::vector<bool> in keep_if_parallelly, fixes #190

### LINKED ISSUE #190: clang osx warning in `find_all_idxs_by`(search.hpp)

Dobiasd:
@pthom Does this fix issue #190 on your system?

### DIFF
--- a/include/fplus/transform.hpp
+++ b/include/fplus/transform.hpp
@@ -400,8 +400,12 @@ typename Container::value_type reduce_1_parallelly(F f, const Container& xs)
 template <typename Pred, typename Container>
 Container keep_if_parallelly(Pred pred, const Container& xs)
 {
+    // Avoid a temporary std::vector<bool>.
     const auto idxs = find_all_idxs_by(
-        identity<bool>, transform_parallelly(pred, xs));
+        is_equal_to<std::uint8_t>(1),
+        transform_parallelly([pred](const auto & x) -> std::uint8_t {
+            return pred(x) ? 1 : 0;
+        }, xs));
     return elems_at_idxs(idxs, xs);
 }
 


pthom:
This fix works perfectly, thanks!

PR #197, opened 2020-11-21
Improve CMake integration

friendlyanon:
This PR reworks the CMake files for this library. For trivial consumption by dependents.  
With this PR one can trivially import this library:
```cmake
# As part of the build tree:
# - add_subdirectory() (e.g. external git clone)
add_subdirectory("${PROJECT_SOURCE_DIR}/external/FunctionalPlus")
# - FetchContent (in CMake 3.14+)
include(FetchContent)
FetchContent_Declare(fplus
        GIT_REPOSITORY https://github.com/Dobiasd/FunctionalPlus
        GIT_TAG ...)
FetchContent_MakeAvailable(fplus)
# As a package installed to the system:
find_package(FunctionalPlus REQUIRED)

# but in all cases the usage is unified
target_link_libraries(my_target PRIVATE FunctionalPlus::fplus)
```
Notice that the dependent isn't required to set any variables beforehand and in all of the cases the header files are included as system headers, which will prevent warnings from being emitted from this library. However, if the library is made part of the build tree, there is still an option to set `FunctionalPlus_INCLUDE_WITHOUT_SYSTEM` to a truthy value to have warnings.

Important things to note:
* Root lists file should be nothing more than build and usage requirements. Dependents should not have to bother with options to turn examples, tests and whatnot off. For this reason there are separate subprojects for tests (`test`), examples (`examples`) and development that loads the entire project for discovery in IDEs (`all`).
* Install rules were not using standard, cross-platform install paths. Those are provided by `GNUInstallDirs`, despite what the name implies.
* Tests' lists file does not use conan anymore. It was only causing conflict and deprives this library from using additional functionality provided by the doctest library's config file, such as test discovery. Yes, I tried using conan, it just broke things.
* The file providing install rules had a name that was suggesting that it did something in relation to [pkg-config](https://en.wikipedia.org/wiki/Pkg-config), which was not the case.
* Toolchain files in relation to CMake are used to set up compiler, linker and such variables crucial for building, but this project was using "toolchain" files which did something completely different.

The commits have a body if the change required an explanation.


Dobiasd:
@friendlyanon Wow, thank you very much! This looks really thought-through. Also, appreciate the accompanying explanation a lot. :+1:

Since the build system is not my main area of expertise in FunctionalPlus, I'll tag a few other contributors, so maybe somebody can help to make a decision. :slightly_smiling_face:

- @ruanformigoni This seems to be related to your [pull request 176](https://github.com/Dobiasd/FunctionalPlus/pull/176). Would the changes here also solve your [issue 175](https://github.com/Dobiasd/FunctionalPlus/issues/175)?

- @danimtb This PR also touches Conan-related things, e.g., `conanfile.py`. Do you see any potential problems with the changes?

- @headupinclouds `pkgconfig.cmake` was initially created by you. What do you think about this PR?

- @pthom Parts of the things you created are also involved, e.g., `toolchain_msvc.cmake`. Do you have an opinion on that?

- @offa `Finddoctest.cmake`, which was originally created by you, no longer exists here. Do you have thoughts about this PR?

friendlyanon:
I just see that I omitted the warnings from the "toolchain" files that provided warning flags, but those were not even in use by CI.

Would you like me to add them back for tests? Currently, there are only two warnings coming from the library:
```
.../include/fplus/container_common.hpp:82:59: warning: unused parameter ‘size’ [-Wunused-parameter]
   82 |     void prepare_container(std::array<T, N>&, std::size_t size)
      |                                               ~~~~~~~~~~~~^~~~
```
```
.../include/fplus/container_properties.hpp:464:10: warning: unused variable ‘size’ [-Wunused-variable]
  464 |     auto size = size_of_cont(xs);
      |          ^~~~
```
Regarding warnings as errors, I think this bit of the video is worth watching: [C++Now 2017: Daniel Pfeifer “Effective CMake"](https://youtu.be/bsXLMQ6WgIk?t=3909)  
But that video in general is worth watching if you have the time for it. It's fuzzy in some places and not entirely correct in a few, but the information in it is already leaps and bounds ahead of most resources out there.

Dobiasd:
Thanks for the remark regarding the warnings. I just [fixed them](https://github.com/Dobiasd/FunctionalPlus/commit/6c7db1310d395ef9f573ec9871312bc11862b06b).

Yeah, I guess it might be helpful to see such warnings in the CI builds.

Also thanks for linking the talk. It looks good. I'll watch it. :+1:
(Up to now, I'm a fan of `-Werror`, because, without proper tooling, it's way too easy to just ignore warnings that not break the build.)

friendlyanon:
I added the warnings (as errors) back for tests and rebased my changes on your fixes.

I accidentally left warnings-as-errors flags commented out and had to push twice, so please cancel the previous CI runs on Travis and Appveyor. Apologies for that.

pthom:
> @pthom Parts of the things you created are also involved, e.g., toolchain_msvc.cmake. Do you have an opinion on that?

I fully agree that what this file was doing is just setting up some warning flags. It is much better to make it a standard cmake script (cmake/warnings.cmake) rather than a toolchain

Dobiasd:
@pthom Thanks. :)

@friendlyanon [pull/176](https://github.com/Dobiasd/FunctionalPlus/pull/176) was just merged into master. It also has some changes in `conanfile.py`, but I guess auto-merging these into your branch should work fine.

friendlyanon:
Yep, Github would let you know on the UI if there was a merge conflict.

Dobiasd:
Just watched the "Effective CMake" talk. Even though I'm still very far from being a CMake pro, I can positively point out, that this pull request, for example, makes us get rid of `add_compile_options`, which is recommended in the ["Forget those commands" section of the talk](https://youtu.be/bsXLMQ6WgIk?t=1104). :grin:

ruanformigoni, danimtb, and pthom already gave their thumbs up, so I'll wait a few more days to see if headupinclouds or offa have something to add. Otherwise, I'll then merge.

CMakeLists.txt
```diff
@@ -1,52 +1,32 @@
...
-  )
+target_include_directories(
+    fplus
+    ${FunctionalPlus_warning_guard}
+    INTERFACE
+    "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
+)
+
+# CMake 3.2 doesn't know cxx_std_14 so instead we use cxx_variable_templates
+target_compile_features(fplus INTERFACE cxx_variable_templates)
```
offa, line 25: Why not increase to a suited CMake Version? :smile: 

@Dobiasd Are there any reasons to stick with v3.2?

all/CMakeLists.txt
```diff
@@ -0,0 +1,97 @@
...
+# Test should be added first, because it disables the warning guard
+add_subdirectory(
+    "${PROJECT_SOURCE_DIR}/../test"
+    "${PROJECT_BINARY_DIR}/test"
+)
+
+# This project supports CMake 3.2 as a minimum version, so there is no way to
+# protect against multiple inclusions other than to turn add_subdirectory into
+# a noop with code injection at project() calls.
+# If at least CMake 3.11 were to be supported, then the FetchContent module has
```
offa, line 16: Ubuntu Bionic ships 3.10 (:cry:) but with Focal it's 3.16 (:+1:).

test/CMakeLists.txt
```diff
@@ -1,124 +1,90 @@
...
+  add_executable("${name}" "${name}.cpp")
+  target_compile_options("${name}" PRIVATE ${project_warnings})
+  target_link_libraries(
+      "${name}"
+      PRIVATE
+      FunctionalPlus::fplus
+      doctest::doctest
+  )
+  # CMake 3.2 doesn't know cxx_std_14 so instead we use cxx_variable_templates
+  target_compile_features("${name}" INTERFACE cxx_variable_templates)
```
offa, line 78: Same as above, maybe we can update CMake?

offa (approved):

offa:
Searching through the CMake doc turned out, `cxx_std_14` is supported since [v3.8](https://cmake.org/cmake/help/v3.8/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#prop_gbl:CMAKE_CXX_KNOWN_FEATURES).

CMakeLists.txt
Dobiasd: Not that I know of. :slightly_smiling_face:

all/CMakeLists.txt
Dobiasd: 3.11. would mean, it would work with Disco Dingo (having 3.12) and up out of the box. For earlier Ubuntu versions one would need to upgrade, which is not too hard with a PPA. So I'd say, let's go with 3.11. (and adjusting [`.travis.yml`](https://github.com/Dobiasd/FunctionalPlus/blob/ec29ec81ab42920967f530a561ef280c1e768a8a/.travis.yml#L3) to `dist: focal`) if this allows implementing things cleaner, and you two don't see a big issue with that. :+1:

offa: Sounds good to me!

Dobiasd (commented):
What do you think about updating to CMake 3.14 then? I'd prefer having a clean implementation even if it means raising the CMake version requirement.

friendlyanon:
Personally I don't see it as necessary, as this is pretty well understood "hack" and there is nothing else that would really require raising the version, especially since you pointed out in https://github.com/Dobiasd/FunctionalPlus/pull/197#discussion_r531162359 that 3.12 would be the highest version you'd raise it to.

Right now the root lists file requires at least 3.8, because of the compile features to set the C++ standard requirement.  
The development lists files require at least 3.11, because of `FetchContent`.

The `ARCH_INDEPENDENT` argument would require both kind of lists files to raise their requirement to 3.14.

Dobiasd:
Ah, sorry, I did not mean "3.12 is the highest version I'd raise to". I understood 3.11 would be nice to have and just wanted to mention that Disco Dingo already supports 3.12, thus 3.11 would not be a problem for me.

If using 3.14 or even 3.16 (or whatever) would help here, let's use this. I expect your changes to be long-lasting, so I'd prefer a long-term solution, even if it means FunctionalPlus appearing a bit "demanding" regarding the CMake version for the first year or so. :slightly_smiling_face:

Dobiasd:
@friendlyanon In case you're waiting for me to greenlight some decision, please let me know. I'd be happy merging this great PR as it, but if you prefer, also after updating to a later CMake version and the accompanying adjustments. I'd trust the judgment of you and @offa on this. :slightly_smiling_face:

friendlyanon:
I guess everyone had enough time to think about it. If you do not mind bumping to 3.14, then I will do that.

Dobiasd:
I don't mind bumping the version. 👍

friendlyanon:
I missed a version when editing a commit, so I am skipping CI run for the commit correcting that single change.

As an aside, you might want to slowly migrate off of Travis CI, as they will essentially be [suspending support for open source projects](https://news.ycombinator.com/item?id=25338983).

Dobiasd:
The commits look good. :+1: Are we ready to merge?

Also thanks for the remark regarding Travis. I'll have a look into GitHub Actions and maybe CircleCI.

friendlyanon:
It seems focal is lacking some g++ versions, I think it would be better to use the Kitware repo for installing the latest CMake version.

I'll revert that commit and make the necessary changes.

friendlyanon:
Well, took a few pushes and then testing in my own repo, but apparently the current Travis configuration for the CI used in this repo prepends a bogus CMake path to `PATH` which is not managed by apt.  
I solved that by just stripping it from `PATH`.

Once the Travis jobs are moved to Github Actions, you can completely drop that part.

Dobiasd:
Thanks a lot. I really appreciate all the effort you put into this. :heart:

I also like that we're still on the positive side of removing lines of code from the repo with this PR. :wink:

YgSLtkl

So when you think we're good to merge, let me know, and I'll do it.

friendlyanon:
Since Travis is green now, I don't have anything else to do here, unless someone has a request for a change.

PR #205, opened 2020-12-08
Timed variadic refs

pthom:
Hello,

I corrected an issue with timed functions and benchmark session.

Basically if you had a function like this, where the second parameter is an output parameter passed by reference:

````cpp
bool foo(int input1, int& output2)
{
    output2 = input1 + 1;
    return true;
}
````

and 
````
auto foo_timed = make_timed_function(foo);
````

Then `foo_timed(a, b)` would not modify the second parameter...

I corrected this by using universal references + std::forward.

> Side note: 
> The C++ syntax for perfect forwarding variadic arguments makes my eyes bleed :-), sorry if had missed that in the first version
> `template<typename ...Args> auto f1()(Args&&... args) { return f2(std::forward<Args>(args)...;) }`
> (oooh, the subtle dance between "...", "&&" and "(" )

### DIFF
--- a/include/fplus/benchmark_session.hpp
+++ b/include/fplus/benchmark_session.hpp
@@ -99,17 +99,17 @@ namespace internal
             , fn_(fn)
         {};
 
-        template<typename ...Args> auto operator()(Args... args)
+        template<typename ...Args> auto operator()(Args&&... args)
         {
-            return _bench_result(args...);
+            return _bench_result(std::forward<Args>(args)...);
         }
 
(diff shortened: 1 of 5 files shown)


Dobiasd:
Oh, good catch! Thanks a lot for fixing, also for the tests. :slightly_smiling_face:

The changes look good to me. Shall I merge?

pthom:
> The changes look good to me. Shall I merge?
Yes!

Issue #203, opened 2020-12-07
conan.io does not show latest version of FunctionalPlus and marks it as being deprecated/legacy


Dobiasd:
On bintray.com, the latest versions of FunctionalPlus released to Conan (`0.2.13` and `0.2.12`) are shown: https://bintray.com/dobiasd/public-conan/functionalplus%3Adobiasd

conan.io, however, only shows the versions up to `0.2.10`: https://conan.io/center/functionalplus

Also, it tells the following:
```
Depreciation Notice: LEGACY PACKAGE
This package may no longer be supported.
```

![1bp66L6](https://i.imgur.com/1bp66L6.png)

@danimtb Did I do something wrong? (I don't see an error in the [upload to Conan in the CI of version `0.2.13`](https://travis-ci.org/github/Dobiasd/FunctionalPlus/jobs/747978815#L1186-L1206) and no big difference to [the past upload of version `0.2.10`](https://travis-ci.org/github/Dobiasd/FunctionalPlus/jobs/730195653).)

Dobiasd:
Yay, it suddenly has the updated version! :tada:

![K2VrK2c](https://i.imgur.com/K2VrK2c.png)

The only thing that still confuses me is: Why is FunctionalPlus marked as legacy/deprecated?

Dobiasd:
Yay, it suddenly stopped showing the deprecation notice. :tada:

![Z64IHp7](https://i.imgur.com/Z64IHp7.png)

The only weird thing now is that it does only show the first part of [the description](https://github.com/Dobiasd/FunctionalPlus/blob/e8285fbc8fe26f40627797e369acfa4c08eb9354/conanfile.py#L7), i.e., instead of

`Functional Programming Library for C++. Write concise and readable C++ code.`

it shows:

`Functional Programming Library for C++.`

But I guess I can live with that. `¯\_(ツ)_/¯`

danimtb:
Hi @Dobiasd,

This is because Conan moved to a "centralized" recipe repository. This means that the community is taking care of creating and maintaining recipes for ConanCenter at https://github.com/conan-io/conan-center-index. This approach makes it easier to add new recipes without adding new files to the repository of the library and lowers the requirements to create packages. This is because the conan-center-index repo provides its own building service that generates the binary packages (or the header only one) automatically when a PR with a recipe is open. Once the PR is merged, packages are promoted to the ConanCenter repo and are publicly available.

[...]

PR #207, opened 2020-12-17
Github Actions

### LINKED ISSUE #206: Migrate to Github Actions?
Travis has changed it's pricing modell ([details](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing)).

Short story: How about moving CI builds to Github Actions?

offa:
CI builds moved to Github Actions (fixes #206). The build shield appears after the first build on `master`.

This PR includes a commit removing Travis CI.

#### Changes compared to Travis CI:

- Update to Focal image
- Older compiler are not included (not supported by Focal without additional effort)
- Building of libc++ no longer necessary (Clang)
- Separate jobs for Linux (GCC / Clang) and OS X (Clang)
- Up-to-date CMake (3.19)

#### To Do:
- [x] GCC 4.9+
- [x] Clang ~~3.7+~~ 4.0+
- [x] Update with `master`
- [x] Doctest fixed version 2.3.4

#### Discussion:

- [x] Conan builds necessary (Fplus builds on conan-center already)? --> no


offa:
Example: https://github.com/offa/FunctionalPlus/actions/runs/428567622

Dobiasd:
Nice! This looks really clean. :+1: 

> Conan builds necessary (Fplus builds on conan-center already)?

I'd love to move the Conan stuff out of the FunctionalPlus repo, but have not yet looked into [it](https://github.com/Dobiasd/FunctionalPlus/issues/203#issuecomment-745211506). Are you interested in that project too? :slightly_smiling_face:

> Should we include older compiler – GCC < 7 / Clang < 8 – again?

If adding them to the `build_config` is as simple as the currently included ones, I think it would make sense to have them there too; at least as long as we claim to support these versions in our README. If not tested, it's too easy to break this compatibility by accident without noticing.

offa:
> I'd love to move the Conan stuff out of the FunctionalPlus repo, but have not yet looked into [it](https://github.com/Dobiasd/FunctionalPlus/issues/203#issuecomment-745211506). Are you interested in that project too?

For me it looks like, there's everything working fine on conan-center – a up-to-date [recipe](https://github.com/conan-io/conan-center-index/tree/master/recipes/functionalplus) (https://github.com/conan-io/conan-center-index/pull/3864) and ready to use [package](https://conan.io/center/functionalplus). I guess we can remove everything conan related (incl. CI) as it's build are maintained over there. It could be a good idea to add a `conanfile.txt` to supply dependencies when building itself  though. Am I right?


> If adding them to the `build_config` is as simple as the currently included ones, I think it would make sense to have them there too; at least as long as we claim to support these versions in our README. If not tested, it's too easy to break this compatibility by accident without noticing.

That's true. Older releases aren't available in the newer ubuntu version (or LLVM PPA), but maybe there's a way to workaround it. I'll have a look at it

Dobiasd:
> For me it looks like, there's everything working fine on conan-center 

Wow, that's nice! I did not expect that, so I did not search for it. But of course, you're right. Yeah, then let's remove the related things from here. :+1:

> It could be a good idea to add a conanfile.txt to supply dependencies when building itself though. Am I right?

I've not really used Conan yet, so I don't know. But if you think it makes sense, I fully believe you. :)

> That's true. Older releases aren't available in the newer ubuntu version (or LLVM PPA), but maybe there's a way to workaround it. I'll have a look at it

Ok, cool. If it adds too much complexity, we can also simply lift our official requirements. GCC 7, for example, is not brand-new, so it's reasonable to require it, I guess.

offa:
I've played around with older GCC and Clang versions, here's what I have so far:

- Clang 7 and 8 are working without much effort
- Clang 6.0 is working too, but needs some extra effort. Broken with _libc++_ (the usual issue); either build from source or use _libstdc++_
- GCC 5 and 6 is not working yet, older version will be difficult

For GCC I'll try a different setup using the official gcc Docker images. This would allow down to 4.9 I guess. Those old versions are EOL, but still available.

There are no official clang images, at least as far as I know. But maybe I can find some suited.

offa:
Using the _gcc_ images it's possible to build gcc-10 to gcc-4.9 without much effort.

There's a good image for _clang_ too, shipping a huge amount of versions. Clang 11 to 7 work fine, but 6.0 to 4.0 with libc++ fail to link due to undefined references; 3.7 doesn't work with libc++ or libstdc++ so I'd suggest dropping that version.

script/ci_build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+mkdir -p build && cd build
+cmake .. -DFPLUS_BUILD_UNITTEST=ON
+cmake --build . --target unittest --config Release -- -j4
```
friendlyanon, line 5: If you want to build the tests, then this should be:
```bash
mkdir build
# Ninja is available in GitHub Actions and automatically uses all cores
cmake -S test -B build -G Ninja
cmake --build build --config Release
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
...
+        if: matrix.build_config.cc == 'clang'
+        run: |
+          sudo script/ci_install_clang.sh ${{ matrix.build_config.version }}
+          if [ "${{ matrix.build_config.version }}" -ge "7" ]; then echo "CXXFLAGS=-stdlib=libc++" >> $GITHUB_ENV; fi
+      - name: Setup
+        run: |
+          sudo script/ci_setup_dependencies.sh
+          sudo script/ci_setup_locales.sh
+      - name: Build
+        run: script/ci_build.sh
```
friendlyanon, line 37: You are not running the tests here. Should have an additional step:
```yml
      - name: Run
        working-directory: build
        run: ctest -C Release -j 4 --output-on-failure
```
Same for the OSX job.

Dobiasd:
Nice investigation!

> Clang 11 to 7 work fine, but 6.0 to 4.0 with libc++ fail to link due to undefined references; 3.7 doesn't work with libc++ or libstdc++ so I'd suggest dropping that version.

Yeah, let's drop it. :sligt_smile:

Btw, do we need AppVeyor for the MSVC tests, or can this (without complications) be done in GitHub Actions too?

script/ci_build.sh
friendlyanon: Ah, I see what happened here, your feature branch is missing the changes from my PR. You should rebase your commits on the current master branch.

Dobiasd:
With [FunctionalPlus now being in conan-center-index](https://github.com/conan-io/conan-center-index/tree/master/recipes/functionalplus), we probably can drop these files:
- `./conan_build.py`
- `./conanfile.py`
- `./test_package/conanfile.py`

Do the following files need some adjustment, or is their Conan part still valid?
- `INSTALL.md`
- `CMakeLists.txt`

README.md
```diff
@@ -1,7 +1,6 @@
 ![logo](logo/fplus.png)
 
 [![CI](https://github.com/Dobiasd/FunctionalPlus/workflows/ci/badge.svg)](https://github.com/Dobiasd/FunctionalPlus/actions)
-[![Build Status Travis](https://travis-ci.org/Dobiasd/FunctionalPlus.svg?branch=master)][travis]
```
Dobiasd, line 3: The following line

`[travis]: https://travis-ci.org/Dobiasd/FunctionalPlus`

can also be removed completely from the README.

script/ci_build.sh
offa: Yes, it's already on the todo list. I'm focusing on getting the environments up first. It great if you could review a second time, once I've merged `master`.

Regarding Ninja: I can add for the CI build, should we use it instead?

README.md
offa: Ah, thanks

script/ci_build.sh
friendlyanon: It's not necessary, just more things to do in the CI which isn't really necessary. If the build was bigger and/or more complex, then maybe Ninja would bring more improvements, but just for the automatic core count usage it's not really worth using over the native tool when you already need the core count anyway to efficiently run `ctest`.

offa:
> Btw, do we need AppVeyor for the MSVC tests, or can this (without complications) be done in GitHub Actions too?

I think so, GH Actions has support for MSVC … however, it's still windows.

offa:
Finally, GCC 10 - 4.9 and Clang 11 - 4.0 are working :tada:. Unfortunately Clang 6 to 4 need to build _libc++_ again, though.

Additionally there are now separate jobs for both compiler.

.github/workflows/ci.yml
offa: _See comments above._

script/ci_build.sh
offa: To get this right, build and test:

```sh
JOBS=4

cmake -S test -B build
cmake --build build --config Release -j $JOBS
ctest -C Release -j ${JOBS} --output-on-failure
```

friendlyanon: Here's the [hardware specification](https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources) for GitHub provided runners if you want to hardcode the core numbers.

offa: @friendlyanon the ctest call mentioned above isn't right I guess:

```
Test project /__w/FunctionalPlus/FunctionalPlus
No tests were found!!!
```

offa: So better pass both – `--config Release` and `-D CMAKE_BUILD_TYPE=Release`?

friendlyanon: `ctest` unfortunately doesn't have a handy directory specifier like `cmake`'s `-B` and `--build`, it has to be run from the build directory, i.e. you need to `cd` first.

offa: I see, adding a `cd` solved the problem.

offa:
All builds ware working now and the branch is rebased onto _master_.

script/ci_setup_linux.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+apt-get update
+apt-get install -y wget git locales
+
+
+## Install CMake
+wget -nv https://github.com/Kitware/CMake/releases/download/v3.19.2/cmake-3.19.2-Linux-x86_64.sh -O setup_cmake.sh
+chmod +x setup_cmake.sh
+./setup_cmake.sh --prefix=/usr/local --skip-license
```
friendlyanon, line 10: This is unnecessary, GitHub provides an up to date CMake in their [virtual environment specifications](https://github.com/actions/virtual-environments).

offa: But how does this get into the docker container?

friendlyanon: `cmake` lives in `/usr/local`, so you mount it:
```yml
container:
  image: ...
  options: -v /usr/local:/host_usr_local
```
and then you make that available in the `PATH`.

offa: Nice, auto update for free!

.github/workflows/ci.yml
```diff
@@ -17,11 +17,14 @@ jobs:
...
     container:
       image: gcc:${{ matrix.build_config.version }}
+      options: -v /usr/local:/host_usr_local
     name: "gcc-${{ matrix.build_config.version }}"
     steps:
       - uses: actions/checkout@master
       - name: Setup
-        run: script/ci_setup_linux.sh
+        run: |
+          echo "/host_usr_local/bin" >> $GITHUB_PATH
```
friendlyanon, line 26: `GITHUB_PATH` is an env variable pointing to a file that you can use to share `PATH` changes across steps in the default runner.

This does not get automatically shared with containers and mounted. It's best to do it the way I commented earlier:
```yml
run: |
  PATH="$PATH:/host_usr_local/bin"
  script/ci_build.sh
```

offa: But then I have to add it to each step using cmake too. Is there a way to set it globally for each job?

offa: According to the [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainerenv), `container` does support an `env`. Is it a bad idea to append the path there?

offa: Odd, even the old GCC / Clang builds use CMake 3.19.1. Maybe the env change is passed to the Image?

friendlyanon: Ah, good to know!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
+name: ci
+
+on: [push, pull_request]
```
friendlyanon, line 3: Might be a good idea to restrict these to run on `master` only. If someone wants to contribute using a feature branch, then the CI won't be redundantly run on both the pushes and the PR.

Dobiasd: Like [so](https://github.com/Dobiasd/enterprython/commit/4cda12f4d735105c5ef8da8eceb00a00a913ae19)? :slightly_smiling_face:

friendlyanon: Yes, exactly.

offa: Sounds good, I'll update the PR.

friendlyanon: Doing this will stop CI from running on feature branches, but PRs targeting the `master` branch will still run as they should.

Not having this filter will run the CI both on the contributor's repository and this repository. Kind of wasteful from an environmental point of view.

offa: But contributors (like me) will get feedback only when opening a PR. Typically PRs are in a mergeable state (or at least close to it) when submitted and don't need many builds within this repository. Removing the feedback for all developers makes a PR the only way to get basic quality checks.

From a developer point of view the quality development and software outweigh a few additional builds. Especially if the builds are within minutes anyway.

Dobiasd: Oh, right. That's also not good. I think I'd [prefer](https://github.com/Dobiasd/enterprython/commit/28d57a2f5656933f304669343dea6acfcc960967) redundant builds over missing builds. It's a bit weird to me that there is no canonical solution to get the best of both worlds.

friendlyanon: Adding `workflow_dispatch` to `on` will enable manual runs. I'd say that is the best of both worlds.

offa: > I'd say that is the best of both worlds.

To be honest: No. Why should one explicitely want to trigger each build manually?

Developing this PR I would have pressed _build_ more than 100 times (literally).

offa: Since I have both of your 👍 I'm going to resolve here 😄 . Thanks for your feedback!

offa:
@Dobiasd ready to merge, I guess.

Dobiasd:
Nice! :tada:

Also, this

ebumhBv

[feels good](https://i.imgur.com/HyM2taZ.png). :)

---

>> Btw, do we need AppVeyor for the MSVC tests, or can this (without complications) be done in GitHub Actions too?

> I think so, GH Actions has support for MSVC … however, it's still windows.

So you think we should leave the MSVC tests on AppVeyor for now?

offa:
> So you think we should leave the MSVC tests on AppVeyor for now?

We can migrate it in an extra step. It might need some extra work.

Dobiasd:
@offa Ok, thanks. Then it's fine from my side. :+1:

@friendlyanon Do you have anything to add?

friendlyanon:
LGTM, AppVeyor can stay as I don't know of them suspending services for open source projects.

Dobiasd:
Yeah, I was more thinking in terms of reducing the overall complexity of the repo by having everything in one place. :slightly_smiling_face:

But independent of that: Great work, @offa. Thanks a lot. :heart:

PR #208, opened 2020-12-18
Simplify "all" source set

friendlyanon:
This will reduce repetition for the defined header list in the `all` subproject using `list(TRANSFORM)`.


Dobiasd:
Yay, functional programming (`TRANSFORM`) in the `CMakeLists.txt` file. :tada: I like it. :grin:

PR #212, opened 2020-12-25
Remove conan

Dobiasd:
Should all be covered here: https://github.com/conan-io/conan-center-index/tree/master/recipes/functionalplus


Dobiasd:
@offa You [mentioned](https://github.com/Dobiasd/FunctionalPlus/pull/207#issuecomment-747617892)

> It could be a good idea to add a conanfile.txt to supply dependencies when building itself though. Am I right?

How should it look like?

offa:
> @offa You [mentioned](https://github.com/Dobiasd/FunctionalPlus/pull/207#issuecomment-747617892)
> 
> > It could be a good idea to add a conanfile.txt to supply dependencies when building itself though. Am I right?
> 
> How should it look like?

It should contain only the dependencies necessary to build fplus locally, eg.:

```
[requires]
doctest/2.4.3

[generators]
cmake_find_package
cmake_paths
```

PR #214, opened 2020-12-29
Move MSVC tests from AppVeyor to GitHub Actions, closes #213

### LINKED ISSUE #213: Move MSVC tests from AppVeyor to GitHub Actions if feasible
This would be nice to reduce the overall complexity of the repository. :)

https://github.com/Dobiasd/FunctionalPlus/pull/214

Dobiasd:


Dobiasd:
@pthom Since you initially created `appveyor.yml`, what do you think of this change?
It allows for quicker feedback, and, I guess, somewhat simplifies the repository. However, I don't get it to use older compiler (`cl.exe`) versions. It's always `The CXX compiler identification is MSVC 19.28.29335.0`.

Dobiasd:
Ah, I found with `runs-on: windows-2016` we get:

```
-- Building for: Visual Studio 15 2017
-- The CXX compiler identification is MSVC 19.16.27045.0
```

And with `runs-on: windows-2019`:

```
-- Building for: Visual Studio 16 2019
-- The CXX compiler identification is MSVC 19.28.29335.0
```

There no `Visual Studio 14 2015`, but maybe we don't need it anymore. :slightly_smiling_face:

offa:
The windows scripts are almost exactly the same as the linux / mac os version. Maybe we can avoid the duplication? Eg. using a shell for the windows run steps.

Dobiasd:
`shell: bash` for Windows is nice. Thanks! Using it I just got rid of the duplicated scripts.

But to make `ci_setup_dependencies.sh` work for Windows too, I needed to do:

- either `(make -j && make install) || (cmake --build . --config Release --target INSTALL)`
- or `if [ $BUILD_USING_CMAKE = true ] ; then ...`

Which one do you think is better? Or do you have a superior solution? :slightly_smiling_face:

pthom:
> @pthom Since you initially created appveyor.yml, what do you think of this change?

Nice job! I think it is much easier to maintain the CI on one platform only (i.e githab actions). Having MSVC 2017 & 2019 is enough (I think we can drop support for MSVC 2015).

offa:
@Dobiasd I haven't tested yet, but is using the cmake call possible on the other systems too?

Dobiasd:
It fails on Linux with `make: *** No rule to make target 'INSTALL'.  Stop.`: https://github.com/Dobiasd/FunctionalPlus/actions/runs/454099641

Dobiasd:
Ah, found the reason. It has to be:

```bash
cmake --build . --config Release --target install
```

instead of:

```bash
cmake --build . --config Release --target INSTALL
```

Now it works fine in all jobs. :slightly_smiling_face:

Dobiasd:
@offa Do you think it's good to merge?

offa:
> @offa Do you think it's good to merge?

I think so! :+1:

PR #216, opened 2021-01-03
CMake install tweaks

friendlyanon:
This contains two Quality of Life improvements:
* Every `install` command that puts files on the filesystem during install now has an associated component with the `${PROJECT_NAME}_Development` naming convention, so they no longer get associated to the default `Unspecified` component.
* If the library is configured as the top level one, the `CPack` module is loaded, which generates `cpack` related files and targets for easy packaging.
```bash
cmake -S . -B build
# no build step, header only project
sudo cmake --install build --component FunctionalPlus_Development
(cd build; cpack -G DEB -D CPACK_PACKAGE_CONTACT=example@example.com)
# FunctionalPlus-0.2.13-Linux.deb generated in build dir, ready to distribute
```
After this PR, I can certainly say that the project could be shown around as doing everything right regarding CMake.

### DIFF
--- a/cmake/install-rules.cmake
+++ b/cmake/install-rules.cmake
@@ -71,4 +71,10 @@ install(
     COMPONENT "${fplus_component}"
 )
 
+# Conditionally include the CPack module only if this is the top project, so
+# appropriate targets and files are generated for packaging the library
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  include(CPack)
+endif()
+
 # }
--- a/cmake/install-rules.cmake
+++ b/cmake/install-rules.cmake
@@ -41,11 +41,16 @@ install(
     INCLUDES DESTINATION "${include_install_dir}"
 )
 
+# Dev component name, useful if the client embeds the library into their build
+# tree and want to install only their own files
+set(fplus_component "${PROJECT_NAME}_Development")
+
 # Headers:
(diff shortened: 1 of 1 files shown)


friendlyanon:
I guess I should elaborate on why adding component names helps other than unclutter the default component.

In UNIX systems' package managers the installable packages are usually split up into multiple components, these are usually data components (eg `cmake-data`), runtime components (eg `libsqlite3-0`) and the corresponding development components (eg `libsqlite3-dev`).  
If you categorize your install components accordingly, it makes it trivial to package a project in a way that abides the conventions of UNIX package management.

Dobiasd:
Thanks!

---

> After this PR, I can certainly say that the project could be shown around as doing everything right regarding CMake.

Yay! :+1:

---

@offa Do you agree with the suggested changes?

offa:
> Every install command that puts files on the filesystem during install now has an associated component with the ${PROJECT_NAME}_Development naming convention, so they no longer get associated to the default Unspecified component.

Is the project name prefix necessary here or optional? If the latter: wouldn't it make sense to drop it and use just a component name like *dev* or *development*? When building fplus the project is obvious and the command becomes `cmake --install build --component development`.

I must admit that I've never used this before, it's more a question of self-interest.

friendlyanon:
Prefixing is necessary, because if the library is pulled into the user's build tree, then you might end up associating your install to theirs. It's also [recommended by Craig Scott](https://youtu.be/m0DwB4OvDXk?t=2168) :)

offa:
Ah, I see! Thanks you!

PR #217, opened 2021-01-06
add install instructions with all in one header

pthom:
Hello, and all the best for this new year!

This PR adds some instructions in install.md concerning the all_in_one header. 

Notes:
1. I also added usage instructions compiler explorer (using the word_stats example we played with a few months ago)
2. I had also considered the possibility to advise to optionally store the last fplus commit SHA1 (for future reference), like this:
````
git ls-remote https://github.com/Dobiasd/FunctionalPlus.git master > external/fplus/GIT_SHA1
````
However, I ended up not adding this in the install.md file.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -175,3 +175,32 @@ brew install functionalplus
 And then, you can add **FunctionalPlus** as a dependency in your cmake project [as in way 1](#cmake-dependency).
 
 If you're not using cmake, you might need to add `$(brew --prefix functionalplus)/include` to the additional include paths for your compiler.
+
+### way 8: using the "all in one" include file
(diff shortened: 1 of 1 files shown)


Dobiasd:
Hi, and happy new year to you too. :slightly_smiling_face:

Thanks, these instructions can be quite helpful. :+1:

I guess you considered to show how to store the last commit SHA for version tracking. But since we also don't do this in the descriptions of the other ways to install using `master`, I guess it's fine without.
An alternative would be to point to a specific release instead, like so: https://raw.githubusercontent.com/Dobiasd/FunctionalPlus/v0.2.13-p0/include_all_in_one/include/fplus/fplus.hpp
But pointing to `master` is fine since it should always be a in usable state.

So from my POV, this PR is ready to be merged. :heavy_check_mark:

pthom:
Also ready to merge for me!

PR #219, opened 2021-01-09
Control doctest from CMake

friendlyanon:
This PR moves config definition for the `main()` implementation from the source files into the tests' lists file.

While I was there, I also corrected the doctest includes to better align with the [core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-incform).

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -67,6 +67,11 @@ set(
 foreach(name IN LISTS tests)
   add_executable("${name}" "${name}.cpp")
   target_compile_options("${name}" PRIVATE ${project_warnings})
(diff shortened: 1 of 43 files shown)


Dobiasd:
Thanks! Moving `DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN` out makes it nicely DRY. :+1:

Regarding the quotes form of include, I probably copy-pasted it from the [doctest tutorial](https://github.com/onqtam/doctest/blob/master/doc/markdown/tutorial.md) back then without thinking about the path not being relative to the cpp file in our case.

PR #221, opened 2021-01-15
Install rules cleanup

### LINKED ISSUE #220: Clean install rules up?
Considering my last sentence in https://github.com/Dobiasd/FunctionalPlus/pull/216#issue-547935166 I feel like that it should be mentioned that many things in the `install-rules.cmake` are remnants of a (not that good) template and I feel like the file should be reworked a bit.

I would remove some comments and variables that aren't really essential, inline the `${PROJECT_NAME}` dereferences (how often does a project change its name?), make code a little tighter, but still preserve the information in the form of comments to stay in the spirit of what I said in that PR.

This issue is to survey whether this change is desired or not.

friendlyanon:
Fixed #220

Please take a look at the comments and let me know if you think they are informative enough without being redundant.

### DIFF
--- a/cmake/install-rules.cmake
+++ b/cmake/install-rules.cmake
@@ -15,8 +15,6 @@ set(generated_dir "${CMAKE_CURRENT_BINARY_DIR}/generated")
 
 # Configuration
 set(version_config "${generated_dir}/${PROJECT_NAME}ConfigVersion.cmake")
-set(project_config "${generated_dir}/${PROJECT_NAME}Config.cmake")
-set(TARGETS_EXPORT_NAME "${PROJECT_NAME}Targets")
 set(namespace "${PROJECT_NAME}::")
 
 # Include module with fuction 'write_basic_package_version_file'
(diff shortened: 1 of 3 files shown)


Dobiasd:
My CMake-fu is way too unsophisticated to review in a meaningful way.

@offa Would you like to have a look? :grin:

offa:
Already :+1:  :-)

cmake/install-rules.cmake
```diff
@@ -1,72 +1,50 @@
...
 )
 
-# Config
-#   * ${CMAKE_INSTALL_LIBDIR}/cmake/FunctionalPlus/FunctionalPlusConfig.cmake
-#   * ${CMAKE_INSTALL_LIBDIR}/cmake/FunctionalPlus/FunctionalPlusConfigVersion.cmake
+# Use the PROJECT_VERSION variable to generate a basic config version file in
+# the PROJECT_BINARY_DIR directory
+write_basic_package_version_file(
+    FunctionalPlusConfigVersion.cmake
+    COMPATIBILITY SameMajorVersion
```
offa, line 27: Maybe we can even use the minor here? @Dobiasd all versions 0.2.x are more or less compatible?

friendlyanon: Is there a need to steer away from semver?

Dobiasd: > @Dobiasd all versions 0.2.x are more or less compatible?

Yeah, they should be. I just went through the changelogs, and I think it was all:
- bug fixes
- new functionality
- performance improvements
- docs
- build system

I can't guarantee 100% backward compatibility, but there were no things breaking it intentionally. So it'd bluntly answer with "Yes.". :slightly_smiling_face:

friendlyanon: For the record, here is the documentation for [write_basic_package_version_file](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file) that explains what the `COMPATIBILITY` argument means.

offa: > Is there a need to steer away from semver?

No, just an idea. It shouldn't block the PR btw.

Dobiasd: Shall we resolve this conversation and merge the PR?

offa: Yes, please.

PR #223, opened 2021-01-20
Update INSTALL.md

friendlyanon:
The lists files for this project changed a lot, so I updated the INSTALL.md file in accordance.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -7,20 +7,17 @@ Requirements and Installation
 You can install FunctionalPlus in **one of the following ways**:
 
 
-### way 1: using [cmake](https://cmake.org/)
+### way 1: using [CMake](https://cmake.org/)
 
 ```bash
 git clone https://github.com/Dobiasd/FunctionalPlus
-cd FunctionalPlus
-mkdir build
-cd build
-cmake ..
-make
-sudo make install
+cmake -S FunctionalPlus -B FunctionalPlus/build
+cmake --build FunctionalPlus/build
+sudo cmake --install FunctionalPlus/build
 ```
 
 <a id="cmake-dependency"></a>
-And then, you can add **FunctionalPlus** as a dependency in your cmake project as in the following.
+And then, you can add **FunctionalPlus** as a dependency in your CMake project as in the following.
 
 ```cmake
 find_package(FunctionalPlus REQUIRED)
@@ -29,14 +26,15 @@ add_executable(HelloWorld main.cpp)
 target_link_libraries(HelloWorld FunctionalPlus::fplus)
 ```
 
-If you want cmake to download and install the package automatically,
+If you want CMake to download and install the package automatically,
 see [ExternalProject](#way-2-using-cmakes-externalproject) below.
 
 #### Building the unit tests
 
-Unit Tests are disabled by default. Building the tests (optional) requires [doctest](https://github.com/onqtam/doctest).
+Unit Tests are available from a separate CMakeLists.txt file. Building the
+tests requires [doctest](https://github.com/onqtam/doctest).
 
-First, install the required locales
+First, install the required locales:
 ````bash
 sudo locale-gen ru_RU
 sudo locale-gen ru_RU.UTF-8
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -132,12 +152,12 @@ cget install Dobiasd/FunctionalPlus
...
 
 
-### way 4: download manually
+### way 5: download manually
 
 Just [download](https://github.com/Dobiasd/FunctionalPlus/archive/master.zip)/extract FunctionalPlus and tell your compiler to use the `include` directory.
 
 
-### way 5: using [Conan C/C++ package manager](https://conan.io)
+### way 6: using [Conan C/C++ package manager](https://conan.io)
```
friendlyanon, line 160: Regarding Conan, this file at the moment suggests using the [`cmake`](https://docs.conan.io/en/latest/reference/generators/cmake.html) generator, however from what I have gathered, the proper generator of choice should be [`cmake_find_package_multi`](https://docs.conan.io/en/latest/reference/generators/cmake_find_package_multi.html), because that generates a real CMake package, however it is marked as experimental and it completely ignores the project's lists files, which already provide a real CMake package.

Can someone comment on this? Can Conan just use the project's CMake files without generating its own files?

Dobiasd:
:heart_eyes:

INSTALL.md
Dobiasd: @danimtb Can you help out here? :slightly_smiling_face:

danimtb: The information here is still valid and the generator to use is up to the user's choice. No need to update anything here.

As a way to clarify about the CMake files: Conan is build-system agnostic and works with the information provided in the conanfile.py package_info(). That information is translated to the build system of choice (generators). That is the reason why packaging CMake files is not needed and even discouraged in OSS projects for better integration with build systems and to avoid some bad practices like hardcoded paths in *Config.cmake/*Find*.cmake files

offa: Why not drop the generator and show the line to add only?

offa: @friendlyanon the package reference has change as it's now available through conan center:

```
functionalplus/0.2.13-p0
```

This PR is a good change to update it :+1: .

INSTALL.md
```diff
@@ -94,35 +89,61 @@ The benefits of this:
...
+include(FetchContent)
+FetchContent_Declare(
+    functional_plus
+    GIT_REPOSITORY https://github.com/Dobiasd/FunctionalPlus.git
+    GIT_TAG master
+)
+FetchContent_MakeAvailable(functional_plus)
+
+add_executable(main src/main.cpp)
+target_compile_features(main PRIVATE cxx_std_14)
```
offa, line 141: Isn't `cxx_std_14` set trhough the fplus target already?

INSTALL.md
friendlyanon: Updated the requires clause.

Another thing I noticed is that `conanfile.txt` is still there, but this project is instead using the `doctest` provided CMake package to be able to use `doctest_discover_tests`.

INSTALL.md
friendlyanon: It's not wise to rely on interface requirements implicitly defining your own requirements.

offa: I see, the example uses C++14 internally, not to get fplus working.

INSTALL.md
offa: > Updated the requires clause.

Perfect, thanks.

 `conanfile.txt` is a way to get development dependencies, eg. instead of cloning the doctest git repository.
Instead of picking an installed library, cmake can pickup the paths shipped by conan.

Dobiasd:
Are the discussions resolved? I.e., can we merge this PR? :slightly_smiling_face:

offa:
I think so.

friendlyanon:
It seems there is a very recent [FreeBSD](https://www.freshports.org/devel/functionalplus) package and there is also a [vcpkg package](), however the vcpkg one is outdated.

Should these also be included? I guess someone can also open an issue at [vcpkg](https://github.com/microsoft/vcpkg) to request an update.

Dobiasd:
Would this add an additional step (update/inform this external package manager), which I have to do manually each time I dare to tag a new release version of FunctionalPlus? :grimacing:

friendlyanon:
Right now, only the Conan instructions have the version number hardcoded. Maybe that one could be made more generic by having a placeholder for the version and instructing the user to consider the latest release when filling in that detail.

I don't really know how consuming a vcpkg library goes, but I know that distribution packages usually don't carry any version info in their install instructions.

Also, I think it's pretty impressive how many places this library is being distributed on.

Dobiasd:
> Also, I think it's pretty impressive how many places this library is being distributed on.

Yeah, I like that too, as long as we don't have to manually take care that those places are updated with new versions. That was my actual concern regarding the addition of a vcpkg explanation. I have no problem with updating a version number in our `INSTALL.md`. I'm scared about needing to jump through several hoops each time by taking care of multiple external package managers to update to (submitting pull requests there, etc.).

As long as a package manager is not listed in our install instructions, I don't care if it has the latest version. But once we have it in there, I somewhat feel responsible for taking care of that.
Conda-forge, for example, picks up new releases automatically. :+1:
Conan does not. :-1:

friendlyanon:
If you want to be cheeky about it: GitHub Actions is really powerful, you could create a workflow that will send PRs/issues/emails to wherever you want to when you push a new tag.

Let the robots do the "dirty" work!

Dobiasd:
Yeah, but this feels like a bit [too much](https://xkcd.com/974/). :grin:

I'm currently more in favor of simply not officially supporting/recommending package managers, which are not as smart as Conda-forge. Also, because `1` package manager has `n` libraries, so I don't see why `n` automations should be developed instead of `1`. :stuck_out_tongue:

friendlyanon:
I think this could be resolved by using the badge generated by repology:  
https://repology.org/project/functionalplus/badges

One issue however is that the FreeBSD package has a different name and the repology response to my merge report was a link to the [package descriptor](https://github.com/freebsd/freebsd-ports/blob/master/devel/functionalplus/Makefile#L3).

I wrote an email to the FreeBSD package maintainer to please look into the possibility of fixing this, so the badge from repology could be used to cover most current and future packages automatically.

offa:
> I think this could be resolved by using the badge generated by repology:
https://repology.org/project/functionalplus/badges

Hehe, I was looking for a conan package badge which includes the version recently. Thanks! :+1:

PR #225, opened 2021-02-27
Fixes / cleanup

offa:
Some minor fixes and dead code removal.


Dobiasd:
Thanks a lot for all of these fixes. :heart:
Regarding the `is_odd_int` error, I'm happy it was "only" in the docs/tests and not in the actual library code. :grimacing:

PR #227, opened 2021-03-11
Broadcast

unixnme:
Hello, thank you for the great library!

I am adding broadcast and zip_broadcast functions.
I think these functions are useful. Not sure if the naming is appropriate.
Please feel free to rename the functions as you see fit.

Somehow, my generated `curry_instances.autogenerated_defines` and `fwd_instances.autogenerated_defines` look quite different from what they should be.
```sh
$ ./generate_derived_functions.sh 
$ head ../include/fplus/curry_instances.autogenerated_defines 
// THIS FILE WAS GENERATED AUTOMATICALLY. DO NOT EDIT.
API search type: identity : a -> a
fwd bind count: 0
--
API search type: is_equal : (a, a) -> Bool
fwd bind count: 1
--
API search type: is_not_equal : (a, a) -> Bool
fwd bind count: 1
--
```

Because of these files, I wasn't able to compile & test
```sh
$ ./compile_all.sh 
In file included from parse_source_files.cpp:7:
In file included from ../include/fplus/fplus.hpp:42:
In file included from ../include/fplus/curry.hpp:85:
../include/fplus/curry_instances.autogenerated_defines:2:1: error: unknown type
      name 'API'
API search type: identity : a -> a
^
../include/fplus/curry_instances.autogenerated_defines:2:11: error: expected
      ';' after top level declarator
API search type: identity : a -> a
          ^
          ;
In file included from parse_source_files.cpp:7:
In file included from ../include/fplus/fplus.hpp:43:
In file included from ../include/fplus/fwd.hpp:119:
../include/fplus/fwd_instances.autogenerated_defines:2:1: error: unknown type
      name 'API'
API search type: identity : a -> a
^
../include/fplus/fwd_instances.autogenerated_defines:2:11: error: expected ';'
      after top level declarator
API search type: identity : a -> a
          ^
          ;
4 errors generated.
./compile.sh: line 9: elm: command not found
./ExploreCompile.sh: line 3: elm: command not found
./TypeSignatureTestCompile.sh: line 3: elm: command not found
./ExploreCompile.sh: line 3: elm: command not found
```

Hope there is no bug in the code.


Dobiasd:
Hi @unixnme, and thanks for the contribution. :slightly_smiling_face:
Could you show a use-case where these two functions are helpful? At first glance, they look very specific to me.
Or could you point to some other functional-programming library providing them?  (The programming language does not matter). This might also help in finding an appropriate name for them. When naming the other functions in FunctionalPlus, I also often looked up if there already is a conventional name for them.

unixnme:
I was thinking application on flattening labeled-grouped data. 
Let's say we have data M that maps team names to their members. For example, M[teamX] == [Ron, Alice], M[teamY] == [Joe, Jane, Ally], ... etc. Then, we want to create a single list of all people and the membership teams, such as 
```
[
 (TeamX, Ron),
 (TeamX, Alice),
 (TeamY, Joe),
 (TeamY, Jane),
 (TeamY, Ally)
 ...
]
```

So, we may do something like
```
apply(M, map_to_pairs(), transform(zip_broadcast()), concat())
```

I may not be aware of a way to do this w/o broadcast.

Dobiasd:
Ah, thanks, got it. Yeah, it makes sense. So

```
zip_broadcast : (a, [b]) -> [(a, b)]
```

would be kind/part of the inverse operation to

```
pairs_to_map_grouped : [(a, b)] -> Map a [b]
```

Since in your use-case you really do start out with a map, how about a function, that actually takes a map as input? E.g.,

```
name_not_yet_decided : Map a [b] -> [(a, b)]
```

Then, instead of something like `apply(M, map_to_pairs(), transform(zip_broadcast()), concat())`, you could simply do the following:

```cpp
name_not_yet_decided(M)
```

Of course, we should check if/how to split it into meaningful and reusable parts. :slightly_smiling_face:

What do you think?

unixnme:
> Since in your use-case you really do start out with a map, how about a function, that actually takes a map as input? E.g.,
> 
> ```
> name_not_yet_decided : Map a [b] -> [(a, b)]
> ```
> 
> Then, instead of something like `apply(M, map_to_pairs(), transform(zip_broadcast()), concat())`, you could simply do the following:
> 
> ```c++
> name_not_yet_decided(M)
> ```
> 
> Of course, we should check if/how to split it into meaningful and reusable parts. 🙂
> 
> What do you think?

Yes, it makes sense to have an explicit function that does `Map a [b] -> [(a, b)]`, since it is the inverse of `pairs_to_map_grouped`. I also think that `zip_broadcast` is a useful part by itself.

For example, consider
```
xs = [[Ron, Alice], [Joe, Jane, Ally], ...]
label = ["teamX", "teamY", ...]
```
In some cases, label may contain duplicate entries, so `pairs_to_map` may not be an option.

In that case, `zip_broadcast` can be useful.
```
apply(xs, zip(label), transform(zip_broadcast()), concat())
```

include/fplus/pairs.hpp
```diff
@@ -151,6 +151,33 @@ auto zip(const ContainerIn1& xs, const ContainerIn2& ys)
...
+        typename Y = typename ContainerIn::value_type,
+        typename ContainerOutX = typename internal::same_cont_new_t<ContainerIn, X>::type,
+        typename ContainerOutY = typename internal::same_cont_new_t<ContainerIn, Y>::type>
+auto broadcast(const X &x, const ContainerIn &ys) {
+    ContainerOutX firsts(ys.size(), x);
+    ContainerOutY seconds{ys.begin(), ys.end()};
+    return std::make_pair(firsts, seconds);
+}
+
+// API search type: broadcast : (a, [b])  -> [(a, b)]
```
Dobiasd, line 169: Changing this to

```cpp
// API search type: broadcast : (a, [b])  -> [(a, b)]
```

should fix your issue with the `autogenerated_defines`.

Dobiasd (changes requested):

include/fplus/pairs.hpp
unixnme: Oops. thank you for catching the typo :) 
It turns out the generate result differed b/c I have BSD `sed` as the default one. After replacing it with `gsed` in the script, the generated file looks as expected.

unixnme:
Testing `apply(xs, zip(label), transform(zip_broadcast()), concat())`, I found a problem here in that `zip_broadcast` should take in a pair<x, ys> for this to work, rather than x and ys as two separate arguments. If we do it, then its argument signature differs from `zip`, where `zip` takes in xs and ys as two separate arguments, whereas `zip_broadcast` takes in a single argument, which is a pair of x and ys. 

So, we would have function signatures 
```
auto zip(const ContainerIn1& xs, const ContainerIn2& ys)
auto zip_broadcast(const std::pair<X, ContainerIn> &pair);
```
which do not seem consistent. Do you have any idea what we can do here to make them more consistent?

In the same logic, should also `broadcast` (if we want to have this as a separate function) take in a single argument as a pair, rather than two separately?

Dobiasd:
Mhh, changing `zip` from taking two containers to taking a pair of two containers is not an option. Not only, because it would be weird to use in many circumstances, but also because it would break a lot of existing client code.

Having `zip_broadcast` taking a value and a sequence to me also looks better than taking a pair of values and a sequence. For the given use-base, maybe instead of using `zip_broadcast` directly, something with `fwd::apply_to_pair(zip_broadcast)` could help?

include/fplus/pairs.hpp
Dobiasd: Oh. Yes, I have to admit the solution using `sed` might not be ideal. :grimacing:

Dobiasd (approved):

Dobiasd:
Instead of `broadcast` and `zip_broadcast`, we could also have just this:

```
zip_repeat : ([a], [b]) -> [(a, b)]
zip_repeat([7,8], [1,2,3,4,5]) == [(7,1), (8,2), (7,3), (8,4), (7,5)]
```

In your case, you would call it with a singleton sequence as the first argument. To me, it currently feels not just a bit more generic, but also a bit saner. What do you think?

unixnme:
> Mhh, changing `zip` from taking two containers to taking a pair of two containers is not an option. Not only, because it would be weird to use in many circumstances, but also because it would break a lot of existing client code.

Oh, yes I agree. I was not suggesting that :)

unixnme:
> Having `zip_broadcast` (and `broadcast`) taking a value and a sequence to me also looks better than taking a pair of a value and a sequence. For the given use-base, maybe instead of using `zip_broadcast` directly, something with `fwd::apply_to_pair(zip_broadcast)` could help?

So I tried it here

```
using namespace fplus;
    using Strings = std::vector<std::string>;
    std::vector<Strings> xs{{"Ron", "Alice"},
                            {"Joe", "Jane", "Ally"}};
    Strings label{"teamX", "teamY"};

    fwd::apply(xs,
               fwd::zip(label),
               fwd::transform(
                       fwd::apply_to_pair(
                               fwd::zip_broadcast()
                       )
               ),
               fwd::concat()
    );
    
    return 0;
```
but it won't compile. I have to make it a lambda function to get it to compile

```
int main() {
    using namespace fplus;
    using Strings = std::vector<std::string>;
    std::vector<Strings> xs{{"Ron", "Alice"},
                            {"Joe", "Jane", "Ally"}};
    Strings label{"teamX", "teamY"};

    fwd::apply(xs,
               fwd::zip(label),
               fwd::transform(
                       fwd::apply_to_pair(
                               [](const auto &x, const auto &y) { return zip_broadcast(x, y); }
                       )
               ),
               fwd::concat()
    );

    return 0;
}
```
which works, but not as pretty and simple.

unixnme:
> Instead of `broadcast` and `zip_broadcast`, we could also have just this:
> 
> ```
> zip_repeat : ([a], [b]) -> [(a, b)]
> zip_repeat([7,8], [1,2,3,4,5]) == [(7,1), (8,2), (7,3), (8,4), (7,5)]
> ```
> 
> In your case, you would call it with a singleton sequence as the first argument. To me, it currently feels not just a bit more generic, but also a bit saner. What do you think?

Yes, this looks good to me.
So, it would be something like
```c++
    auto result = fwd::apply(xs,
                             fwd::zip(label),
                             fwd::transform(
                                     fwd::transform_fst(fwd::singleton_seq())
                             ),
                             fwd::transform(
                                     fwd::zip_repeat()
                             ),
                             fwd::concat()
    );

    return 0;
```

How about `name_not_yet_decided : Map a [b] -> [(a, b)]` function? Will we still have this as a separate function?

Dobiasd:
How about `name_not_yet_decided : Map a [b] -> [(a, b)]` function? Will we still have this as a separate function?

Yeah, I still think it makes sense because your use-case looks reasonable and not too rare to me. Just to make sure we mean the same thing, I was thinking about something like this (with your use-case in the test):

```cpp
// API search type: name_not_yet_decided : Map key [val] -> [(key, val)]
// fwd bind count: 0
// Convert a dictionary with sequence values to a list of key-value pairs.
// Inverse operation of pairs_to_map_grouped.
// name_not_yet_decided({"a": [1, 2, 4], "b": [6]})
//     -> [("a", 1), ("a", 2), ("b", 6), ("a", 4)]
template <typename MapType,
    typename ContainerOut = std::vector<
        std::pair<
            typename MapType::key_type,
            typename MapType::mapped_type::value_type>>>
ContainerOut name_not_yet_decided(const MapType& dict)
{
    ContainerOut result;
    for (const auto& key_and_values : dict)
    {
        for (const auto& value : key_and_values.second)
        {
            result.push_back(std::make_pair(key_and_values.first, value));
        }
    }
    return result;
}

TEST_CASE("maps_test, name_not_yet_decided")
{
    using namespace fplus;
    std::map<std::string, std::vector<std::string>> teams = {
        {"teamX", {"Ron", "Alice"}},
        {"teamY", {"Joe", "Jane", "Ally"}}
    };
    // This is your use-case, right?
    const auto result = name_not_yet_decided(teams);
    REQUIRE_EQ(result.size(), 5);
}
```

Of course, `name_not_yet_decided` not only needs a name, but it also could be implemented by using smaller (reusable) functions (like you suggested) instead of imperatively like in the ad-hoc implementation above.

unixnme:
will open a separate PR for name_not_yet_decided : Map a [b] -> [(a, b)]
closing this, as broadcast/zip_broadcast won't be included

Issue #184, opened 2020-09-21
Usage instructions


jchia:
How is the user supposed to use fplus? Is including fplus/fplus.hpp the only way? Can he include other non-internal headers instead? If so, what combinations of other headers are allowed? This question is necessary because currently, certain combinations break, e.g. just including fplus/maybe.hpp breaks but including fplus/container_common.hpp followed by fplus/maybe.hpp is fine. (Also see #183.) There are circular dependencies between the headers, so just adding includes until the compiler shuts up does not always work because the order of include matters.

If only fplus/fplus.hpp is supposed to be included, I think the README should say so clearly, especially because of the existence of the internal/ directory implying that non-internal things can be used. Including fplus/fplus.hpp is not cheap, though, and costs 0.5s more to compile on my machine than just including fplus/container_common.hpp (0.77s vs 0.27s) .

If other headers can also be included by the user instead of fplus/fplus.hpp, then I think the rules for including them should be specified in the README. All the examples in the README just include fplus/fplus.hpp.

Dobiasd:
Hi @jchia,

thanks a lot. Of course, you're totally right. Ideally, it would be possible to include each header not in the `internal` directory in isolation. I'll check if it is feasible to achieve this state, and also try to get rid of the circular includes.

If this is not possible, I'll adjust the `README.md` so that it explains, that including `fplus/fplus.hpp` is the only way.

Dobiasd:
So, I've [fixed the includes](https://github.com/Dobiasd/FunctionalPlus/commit/4260e3915c0c15a578f6245bd1fe9b397993a255). :slightly_smiling_face:

Now every non-internal `.hpp` file can be included separately.

Also I've made sure no circular include dependencies are left, which I can be verified as follows:

```bash
cd include
echo "import networkx as nx" > detect_cycles.py
echo "edges = [" >> detect_cycles.py
grep -r . -e "#include" | perl -pe "s/^\.\///g" | perl -pe "s/.#include </,/g" | perl -pe "s/>$//g" | perl -pe "s/\//_/g" | perl -pe "s/\./_/g" | perl -pe "s/^/('/g" | perl -pe "s/,/','/g" | sed "s/$/'),/g" | grep -v autogenerated  >> detect_cycles.py
echo "]" >> detect_cycles.py
echo "cycles = list(nx.simple_cycles(nx.DiGraph(edges)))" >> detect_cycles.py
[...]

PR #228, opened 2021-03-13
Zip repeat

unixnme:
I am adding zip_repeat, as you have suggested in #227 .


Dobiasd:
Thanks a lot. In general, it looks good to me. The implementation is not full of `if`s and nicely symmetric. :+1:  The tests are also nicely concise. :slightly_smiling_face:

I'm just a bit worried about the need for the forward declaration. As far as I remember, this was not needed up to now. Do we have a problem with our include-dependency tree?

unixnme:
> I'm just a bit worried about the need for the forward declaration. As far as I remember, this was not needed up to now. Do we have a problem with our include-dependency tree?

With `zip_repeated`, yes. We now have generate.hpp -> numeric.hpp -> pairs.hpp -> generate.hpp.
Options that I can think of are
1. forward declare
2. implement zip_repeat w/o using existing repeat
3. move `repeat` to a different header, maybe container_common?

Any idea / suggestion on this?

Dobiasd:
Ah, yeah. I just checked out your branch

```bash
git clone https://github.com/Dobiasd/FunctionalPlus
cd FunctionalPlus
git fetch origin pull/228/head:zip_repeat
git checkout zip_repeat
```

and checked for circular includes like [so](https://github.com/Dobiasd/FunctionalPlus/issues/184#issuecomment-698169328), and it printed:

```
1 cycles found:
['fplus_numeric_hpp', 'fplus_pairs_hpp', 'fplus_generate_hpp']
```

We should avoid having circles. :)

---

Regarding your three options:

1. Meh.
2. Would not be beautiful, but ok.
3. Would be ok too. Even users who did not `#include <fplus/fplus.hpp>` but only `#include <fplus/generate.hpp>` and use `repeat` would still be ok, since `generate.hpp` includes `container_common.hpp`. Actually, I did this move multiple times in the past. Many things in `container_common.hpp` are just there because I tried to avoid cycles. :grin:

But in this case, I think there is a fourth, even simpler, solution. I think `generate.hpp` does even need to include `numeric.hpp`. So removing that include should fix the problem. :slightly_smiling_face:

unixnme:
> But in this case, I think there is a fourth, even simpler, solution. I think `generate.hpp` does even need to include `numeric.hpp`. So removing that include should fix the problem. 🙂

fantastic! Fixed now 👍 
Not sure why my IDE was not able to flag it as unused #include

Dobiasd:
:tada:

From my point of view, we are ready to merge. What do you think?

unixnme:
> 🎉
> 
> From my point of view, we are ready to merge. What do you think?

looks good to me too!

Dobiasd:
:+1:

I've updated the [API search](http://www.editgym.com/fplus-api-search/) accordingly:

x82Mf32

PR #229, opened 2021-03-15
Grouped map to pairs

unixnme:
Hello,

I added `map_grouped_to_pairs` function as we have discussed in #227 .
I kept your description, but removed ContainerOut template (for simplicity) and re-implemented using existing functions.


Dobiasd:
Thanks! I like the elegant implementation. :heart_eyes:

The restoration test (`x |> map_grouped_to_pairs |> pairs_to_map_grouped`) is also a good idea. :+1:

unixnme:
but it's failing the ci test. :(
it works fine on mine (macOS, clang12)

Dobiasd:
Weird. It does look correct though. I guess it's just too much for some type-deduction implementations. Basically, we're running into compiler bugs.

Replacing

```cpp
        return apply_to_pair(zip_repeat<std::vector<Key>, Group>,
                             transform_fst(singleton_seq<Key>, pair));
```
with

```cpp
        const auto f = zip_repeat<std::vector<Key>, Group>;
        return apply_to_pair(f, transform_fst(singleton_seq<Key>, pair));
```

should do the trick. (At least it does on my machine.)

Dobiasd:
Looks better. :+1:

If you move that one `{` to the next line, I guess we're good to merge.

unixnme:
> Looks better. 👍
> 
> If you move that one `{` to the next line, I guess we're good to merge.

Done :)

Dobiasd:
Thanks.

Also, in general. It is very pleasant discussing things with you and seeing you using fplus so aptly. :-)

PR #230, opened 2021-03-15
Restore include_all_in_one build

pthom:
Hello,

I saw that the generation of the file `include_all_in_one/include/fplus/fplus.hpp` is not included anymore in the make step 

This file is part of the repository so that it is easily accessible via an URL, for easy install via only one file, or for use on tools such as compiler explorer.

I suspect this happened during the merge of the PR "[Improve CMake integration](https://github.com/Dobiasd/FunctionalPlus/pull/197)", since this PR vastly refactored the cmake scripts. 

So, I open this PR as a way to discuss what can be done on this subject:
 
- This PR proposes to restore the original status: this file is generated automatically at each build. The advantage is that developers will not forget to commit it, since it is refreshed automatically for them

- However, we could imagine less intrusive ways to do it: either by mentioning inside [CONTRIBUTING](https://github.com/Dobiasd/FunctionalPlus/blob/master/CONTRIBUTING.md) the need to call manually `cd include_all_in_one && python3 make_all_in_one.py && cd ..` for each PR, or by calling this script during other deploy scripts (such as `compile_all_and_deploy.sh` or `generate_derived_functions.sh`)

What do you think? 

Note: this PR also update the `include_all_in_one/include/fplus/fplus.hpp` with the latest modifications.


Dobiasd:
Oh! :grimacing: Thanks a lot for noticing and reporting. :slightly_smiling_face:

I wonder if it would make sense to not only do this auto-generation step automatically when building, but also the fwd/curry stuff. WDYT?

pthom:
> I wonder if it would make sense to not only do this auto-generation step automatically when building, but also the fwd/curry stuff. WDYT?

I think it would be a good idea! 
I would be interested in giving this a try. Would you like me to propose something around this in this PR?

CMakeLists.txt
```diff
@@ -29,3 +29,7 @@ target_link_libraries(fplus INTERFACE Threads::Threads)
 # ---- Install ----
 
 include(cmake/install-rules.cmake)
+
+# ---- Make amalgamated library ----
+
+add_subdirectory(include_all_in_one)
```
friendlyanon, line 35: Should go betwen the install rules and the library declaration.

include_all_in_one/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# This script will create the amalgamated library into
```
friendlyanon, line 1: Please inline this into the root CML file, prefer not to fragment it if not necessary. When inlining, follow the style existing style, i.e. create a new section like `# ---- Declare include_all_in_one ----`

include_all_in_one/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+#    include_all_in_one/include/fplus/fplus.hpp
+#
+# Notes:
+# - include_all_in_one/include/fplus/fplus.hpp is part of the source repository, so that using FunctionalPlus 
+#   online with tools such as Compiler Explorer is easy. Just write:
+#     #include "https://raw.githubusercontent.com/Dobiasd/FunctionalPlus/master/include_all_in_one/include/fplus/fplus.hpp"
+# - this file will be updated automatically at each build, so that library developpers do not forget 
+#   to commit it (ideally once per PR)
+find_package (Python COMPONENTS Interpreter QUIET)
+if (Python_FOUND)
```
friendlyanon, line 11: Please format the commands in a way that is consistent with the rest of the project:

* No space after the command's name
* 2 spaces to indent
* 4 spaces for continuation

include_all_in_one/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+    add_custom_target(
+        include_all_in_one
+        ALL
+        COMMAND ${Python_EXECUTABLE} ${CMAKE_CURRENT_LIST_DIR}/make_all_in_one.py
```
friendlyanon, line 15: Please always quote arguments that have variable expansion in them. Unquoted variable expansion can result in unintended list expansion, but this is also to be consistent with the rest of the files.

Once this is inlined into the root CML, please use `PROJECT_SOURCE_DIR` instead of `CMAKE_CURRENT_LIST_DIR` for consistency.

Since this command generates a file that is known ahead of time, the `BYPRODUCTS` argument should be used to indicate as such.

include_all_in_one/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+    )
+else()
+    message(STATUS "Not generating include_all_in_one, python was not found.")
```
friendlyanon, line 18: The message isn't quite right. This command adds a target that when built generates the all in one, it doesn't generate it by itself. The more correct message would be `Python not found, not adding the 'include_all_in_one' target`

friendlyanon (changes requested):
Just a few changes to make it fit better with the existing structure of the project CML.

CMakeLists.txt
```diff
@@ -26,6 +26,21 @@ target_compile_features(fplus INTERFACE cxx_std_14)
 find_package(Threads REQUIRED)
 target_link_libraries(fplus INTERFACE Threads::Threads)
 
+# ---- Create amalgamated library ----
+
+# This will create the amalgamated library into include_all_in_one/include/fplus/fplus.hpp
+find_package(Python COMPONENTS Interpreter QUIET)
+if (Python_FOUND)
```
friendlyanon, line 33:
```suggestion
if(Python_FOUND)
```

Since `if` is "just" a command, there should be no space after it.  
All good after this.

pthom:
@friendlyanon : I took into account your suggestions. I also added a call to `generate_derived_functions.sh`with the same philosophy. 

@Dobiasd : `generate_derived_functions.sh` is now called at each build. However this will work under linux and MacOS, but not windows. If you look and its code, you will see that  I had to adapt it for MacOS (where sed is sadly not GNU compatible).
You will also see that I tried to adapt the contributing.md doc.

CMakeLists.txt
```diff
@@ -26,6 +26,34 @@ target_compile_features(fplus INTERFACE cxx_std_14)
...
 target_link_libraries(fplus INTERFACE Threads::Threads)
 
+# ---- Create auto-generated fwd and curried functions ----
+
+if(UNIX)
+  add_custom_target(
+      generate_derived_functions
+      ALL
+      COMMAND bash "${PROJECT_SOURCE_DIR}/generate/generate_derived_functions.sh"
+      BYPRODUCTS "${PROJECT_SOURCE_DIR}/include/fplus/fwd_instances.autogenerated_defines" "${PROJECT_SOURCE_DIR}/include/fplus/curry_instances.autogenerated_defines"
```
friendlyanon, line 36: To keep things on the left nice and tidy within 80 columns, `BYPRODUCTS` could go on its own line, followed by the byproducts on each new line.

friendlyanon (approved):
The CMake side of things look perfect to me.

Dobiasd:
Perfect, thanks a lot! (Also for adjusting the docs. :heart:) I've just checked out your branch locally, and the auto-generation during build works fine here too. :+1:

So I guess we're ready to merge, right?

For the long-term, we could think about if we'd like to replace my weird `cat | grep | perl | sed` chain with something less insane. :grin:

friendlyanon:
Does the shell script use the compiler just for resolving the `#include`s? If yes, then you can do the same with just a Python script and that's be cross-platform as well.

generate/generate_derived_functions.sh
```diff
@@ -1,11 +1,28 @@
 #!/usr/bin/env bash
 
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+cd $THIS_DIR
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
xtofl, line 6: Just an idea: this is a switch.  It is disguised as an `if/else` block.  Wouldn't it be showing intent more clearly with

```bash
function assert_gsed_available() {
   if ! command -v gsed &> /dev/null; then
        echo "gsed could not be found (gsed is the GNU implementation of sed, required by this script, under MacOS)
        Please install it with:
            brew install gnu-sed
        "
        exit 1
    fi
}


case "$OSTYPE" in
  darwin)
      assert_gsed_available || exit 1
      SED_CMD=gsed
      ;;
  *) SED_CMD=sed;;
esac

**Comment by Dobiasd** [comment]
@friendlyanon

> Does the shell script use the compiler just for resolving the `#include`s?

Yes, it does:

```bash
g++ -E -CC -std=c++14 -I../include ../include/fplus/fplus.hpp > temp_preprocessor_output.txt
cat temp_preprocessor_output.txt | ...
```

> If yes, then you can do the same with just a Python script and that's be cross-platform as well.

Yeah, that would be better. :+1:

generate/generate_derived_functions.sh
pthom: Good idea, I did this

pthom:
I added some more commits, which summarize to:

#### [Group derived and amalgamation targets into one target ("auto_generate")](https://github.com/Dobiasd/FunctionalPlus/pull/230/commits/633511c5b40f52d65a213e36fa0ca87dd76c5f08)

we do not use 'gcc -E' anymore for derived function script; instead we use the result of the make_all_in_one.py script

Timing results:

After this commit (only one "auto_generate" target)
```
> time make
Built target auto_generate
make  0.15s user 0.13s system 116% cpu 0.329 total
````

Before this commit (two targets)
```
> time make
Built target generate_derived_functions
-- Generating include_all_in_one
Built target include_all_in_one
make  0.41s user 0.13s system 111% cpu 0.488 total
````
 
#### [Refactor and document auto_generate.sh](https://github.com/Dobiasd/FunctionalPlus/pull/230/commits/dc308414a828d19ab1e31b140877f361ba0cc295)

Here, I refactored the bash code generation part, in order to prepare a possible migration to python.

The heart of the algorithm is here:

````bash
function perform_code_generation () {
    if [ "$#" -ne 3 ]; then
        echo "perform_code_generation: illegal number of parameters"
        exit 1
    fi
    dst_filename=$1
    search_token=$2
    replace_token=$3

    cat $ALL_IN_ONE_FILE \
    | grep -B 1 -e "// $search_token" \
    | perl -pe "s/^...//g" \
    | $SED_CMD ':a;N;$!ba;s/\n/ /g' \
    | perl -pe "s/ -- /\n/g" \
    | perl -pe "s/API search type: ([^ ]*).*fwd bind count: ([0-9])/$replace_token\2(\1)/g" \
    >> $dst_filename
}
````

However, I have a hard time deciphering the sed part; so that I stopped here.

**Comment by pthom** [comment]
Ok, give me some time to finish the port to python :-)

**Comment by friendlyanon** [comment]
For the record the sed script is just:
```sh
:a       # create label a
N        # append input to pattern space
$!ba     # branch to label a on every line but the last one
s/\n/ /g # replace \n with a space in the pattern space
```

pthom:
Ok, the generate part is now full python. The code is heavily documented and should be (I hope) quite readable.
See [here](https://github.com/Dobiasd/FunctionalPlus/pull/230/files#diff-3e94ff4bb2ac3588a475b4bbc98d588820afc2f3cfb9a16bc1f2db7ccb6d49f5)

Do you want me to rewrite this PR history in one or two commits?

Dobiasd:
> Ok, the generate part is now full python.

Awesome! :rocket: :rocket: :rocket:

> The code is heavily documented and should be (I hope) quite readable.

Yes, I think so too. :+1:
The `sed` version was very terse, but unreadable "write-only code". :grimacing:

> I also updated the doc.

:heart:

> Do you want me to rewrite this PR history in one or two commits?

I don't have a strict policy regarding the commit history for this repository. But if you'd like your contributions to look neater, feel free to rewrite the history. :slightly_smiling_face:

friendlyanon:
> The `sed` version was [...] "write-only code"

I wouldn't say that, took me one or two minutes cross-referencing the man pages to turn it back into english. It just comes down to familiarity, someone who does a some more sed scripting will know it right away, as sed doesn't have a huge number of commands, so it's relatively easy to keep in mind.

pthom:
OK, I rewrote the history into one commit. 
If everyone agrees, it is ok to merge IMHO.

PR #237, opened 2021-04-05
Remove string-locale tests

Dobiasd:


Dobiasd:
Currently, I favor removing these tests completely in favor of [re-enabling them](https://github.com/Dobiasd/FunctionalPlus/pull/236), because it simplifies building/running the tests.

(The two functions under test, i.e., [`fplus::to_lower_case_loc`](https://github.com/Dobiasd/FunctionalPlus/blob/cad25652fdc9fdc82ca6f5bc885b749e808591a3/include/fplus/string_tools.hpp#L127-L136) and [`fplus::to_upper_case_loc`](https://github.com/Dobiasd/FunctionalPlus/blob/cad25652fdc9fdc82ca6f5bc885b749e808591a3/include/fplus/string_tools.hpp#L127-L136) are very simple anyway.)

PR #240, opened 2021-04-06
Fix #239: auto_generate.py touches nothing when no change is detected

### LINKED ISSUE #239: Consecutive build should only re-build changed files instead of all. (Circular build dependency in CMake setup?)
```bash
git clone https://github.com/Dobiasd/FunctionalPlus
cmake -S FunctionalPlus/test -B FunctionalPlus/build
cmake --build FunctionalPlus/build -j 8
cmake --build FunctionalPlus/build -j 8
cmake --build FunctionalPlus/build -j 8
```

The first call to `cmake --build FunctionalPlus/build` is supposed to build all (`Building CXX object CMakeFiles/foo_test.dir/foo_test.cpp.o`), but the second and third ones should not do anything. Instead, they rebuild everything. The superfluous waiting time can be annoying when locally modifying only one of the unit tests.

The custom target `auto_generate` (added with [pull/230](https://github.com/Dobiasd/FunctionalPlus/pull/230)) produces `include/fplus/fwd_instances.autogenerated_defines` and `include/fplus/curry_instances.autogenerated_defines`, which are inputs for the unit tests.

`auto_generate` depends on `ALL`. Does that mean, that it is also triggered (overwrites the `autogenerated_defines` files) when the tests were built? (Then, the next build needs to re-build the tests, and the cycle continues.)

If so, I guess it would make sense to run `auto_generate` before (and independent of) the tests to break the cycle. This would make additional sense because the tests actually do depend on the content of the `autogenerated_defines` files but not vice versa.

/cc @friendlyanon @pthom

pthom:
The code should be self explanatory: if old_content == content, then auto_generate should not touch the generated files (and thus, it cannot trigger rebuilds).


Dobiasd:
Thanks for the quick help. That fixes the issue of superfluous re-builds. :+1:

I just tested:

```bash
git clone https://github.com/Dobiasd/FunctionalPlus
cd FunctionalPlus
git fetch origin pull/240/head:fix_circular_generation
git checkout fix_circular_generation
cd ..
cmake -S FunctionalPlus/test -B FunctionalPlus/build
cmake --build FunctionalPlus/build -j 8
cmake --build FunctionalPlus/build -j 8
cmake --build FunctionalPlus/build -j 8
```

(Build 2 and 3 are now very fast, as they should be.)

---

But the second problem from #239 persists, i.e., `auto_generate` produces the `.autogenerated_defines` files. The tests use them (via `#include`), but this dependency is not reflected in our CMake configuration yet. The tests should depend on `auto_generate`.

With the current config (test not depending `auto_generate`), the following situation happens:
- You implement a new function `fplus::foo`.
- You write a test for `fplus::fwd::foo`.
- You try to build.
- The tests don't compile. (`error: foo is not a member of fplus::fwd`)

Now, depending on if `auto_generate` was already started when the above compilation error occurred (when building with `-j`), the new `autogenerated_defines` might already have been produced or not. If they were, the next attempt to build will succeed. If not, you need to try again until `auto_generate` was run in parallel to the error-generating compilation (and thus finished), or you somewhat need to run it manually.

I guess once the dependency graph is fixed, `auto_generate` maybe would not even need to check if the output is different from last time, because it would not event be invoked when its input did not change.

test/CMakeLists.txt
```diff
@@ -75,6 +75,11 @@ foreach(name IN LISTS tests)
       FunctionalPlus::fplus
       doctest::doctest
   )
+
+  if(TARGET auto_generate)
+    add_dependencies("${name}" auto_generate)
```
friendlyanon, line 80: This should go in the main CML unconditionally after the custom target.

pthom:
> But the second problem from #239 persists, i.e., auto_generate produces the .autogenerated_defines files. The tests use them (via #include), but this dependency is not reflected in our CMake configuration yet. The tests should depend on auto_generate.

The second commit introduces a dependency: tests depend on auto-generate, when this target is available (i.e if python is found).

> I guess once the dependency graph is fixed, auto_generate maybe would not even need to check if the output is different from last time, because it would not even be invoked when its input did not change.

Remarks on the input depencies of `auto-generate`:

* The `ALL` parameter of [add_custom_target](https://cmake.org/cmake/help/latest/command/add_custom_target.html) does not indicate dependency per se, it indicates that this target should be added to the default build target so that it will be run every time.

* If we really want to add input dependencies to auto-generate, we should use the `DEPENDS` param of add_custom_target. However, basically this would imply to list there all the fplus hpp files (since basically the fwd files do depend on all of them). This is why I preferred to use `ALL`. 

However if you accept using GLOB (which is considered heretic by some), this could be solved with not to much copy/pasta.

test/CMakeLists.txt
pthom: > This should go in the main CML unconditionally after the custom target

We are inside a foreach loop inside tests/CMakeList that produces multiple targets which are unknown to the main CMakeLists. I do not see an obvious way for how this could be done in the main  CML

Dobiasd:
Ah, thanks for the explanation. :slightly_smiling_face:

Ok, to avoid manual listing of inputs and to avoid doing heretic things, I guess for me it's ok like you're doing right now (in this PR). `auto_generate` will be run every time, but it is quite fast, so it does not take long until it realizes that it does not need to overwrite its output.

pthom:
> Edit: I've verified that it does the right thing by looking at the output of this (expectations described in the comments).

I had also checked that adding a `fplus::my_foo_function` and a test for `fplus::fwd::my_foo_function`now works on the first build.

test/CMakeLists.txt
friendlyanon: `add_dependencies(fplus auto_generate)`

pthom: Ok, thanks & done ;-)

PR #241, opened 2021-05-04
GCC 11 and Clang 12 builds

offa:
CI build for the latest versions of GCC and Clang.


Dobiasd:
Nice, thank you! :heart:

PR #244, opened 2021-08-01
Avoid self-assignment in keep_by_idx, fixes #243

### LINKED ISSUE #243: keep_by_idx() returns list of empty elements.
`vector<string> const strings{"str1", "str2"};`                                                             
`auto result = fplus::keep_by_idx([](auto idx) { return true; }, strings);`       

The result should be {"str1", "str2"}, but it is actually {"", ""}.
I think line 114 in filter.hpp has something to do with the problem.

Dobiasd:
Fixes #243.

### DIFF
--- a/include/fplus/container_common.hpp
+++ b/include/fplus/container_common.hpp
@@ -138,6 +138,13 @@ namespace internal
         return std::back_inserter(ys);
     }
 
+    // Avoid self-assignment.
+    template <typename T>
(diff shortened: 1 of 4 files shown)


Dobiasd:
@lizhuimu What do you think about this solution? :slightly_smiling_face:

Dobiasd:
Ok, I interpret your reaction emoji (:heart:) as an approvement of this pull request. :wink:

PR #245, opened 2021-08-31
Update GH Action branches

offa:
The `checkout` step has switched to `main` as default branch. `master` still exists, but isn't updated anymore.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       options: -v /usr/local:/host_usr_local
     name: "gcc-${{ matrix.build_config.version }}"
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@main
       - name: Setup
         run: |
           echo "/host_usr_local/bin" >> $GITHUB_PATH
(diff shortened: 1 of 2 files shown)


Dobiasd:
Should we also rename the branch `master` to `main`? It looks like currently it still has the old name:
- https://github.com/Dobiasd/FunctionalPlus/tree/master
- https://github.com/Dobiasd/FunctionalPlus/tree/main -> `404`

offa:
While `main` is the new default on GH, `master` is still a valid default branch.
Changing the name is purely a personal decision of the maintainer – there's no technical reason.

Dobiasd:
Ok, thanks. Do you expect any troubles after renaming? Otherwise, I guess I might do it just for the sake of consistency.

offa:
> Do you expect any problems after renaming? 

I don't think so, at least regarding this repo. Only some docs need an update: https://github.com/Dobiasd/FunctionalPlus/search?q=master

However, users and forks need to adopt too. It might be a good idea to keep master for some time and create a pinned issue. Even though master wont get any updates, it gives users some time.

Dobiasd:
Good idea, I just did so:
- https://github.com/Dobiasd/FunctionalPlus/issues/246
- https://github.com/Dobiasd/frugally-deep/issues/302

Also, I've ported your change from this PR over to frugally-deep:
- https://github.com/Dobiasd/frugally-deep/pull/301

Thanks! :heart:

offa:
Perfect! :tada:

PR #247, opened 2021-09-22
Fix is_odd function. Added unit tests to verify both is_even and is_odd.

moowlf:
I believe i found an error in the function [id_odd](https://github.com/Dobiasd/FunctionalPlus/blob/aa24d89ec24a412d284bc25d36c1ff40285dd17c/include/fplus/numeric.hpp#L138) in numerical.hpp.

### DIFF
--- a/include/fplus/numeric.hpp
+++ b/include/fplus/numeric.hpp
@@ -138,7 +138,7 @@ template <typename X>
 bool is_odd(X x)
 {
     static_assert(std::is_integral<X>::value, "type must be integral");
-    return x % 1 == 0;
+    return x & 1;
 }
 
 namespace internal
--- a/include/fplus/numeric.hpp
+++ b/include/fplus/numeric.hpp
@@ -138,7 +138,7 @@ template <typename X>
 bool is_odd(X x)
 {
     static_assert(std::is_integral<X>::value, "type must be integral");
-    return x & 1;
+    return x % 2 != 0;
 }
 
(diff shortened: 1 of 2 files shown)


Dobiasd:
Oh, thanks a lot! :slightly_smiling_face: :+1:

It's somewhat embarrassing, I probably copied `is_odd` from `is_even` (`x % 2 == 0`) and then wanted to replace the `0` with a `1` (`x % 2 == 1`), but instead replaced the `2` with a `1` (`x % 1 == 0`). :grimacing: In addition, I felt it's so simple, I don't need to write tests for it (also wrong). :woozy_face:

---

Is `x & 1` the same as `x % 2 == 1` in all cases?

Dobiasd:
Ah, no, for negative inputs, only `x & 1` works correctly. Interesting. Would you like to add a unit test for negative input too (to avoid somebody accidentally change it to `x % 2 == 1` in the future)?

Then, I think, we'd be ready to merge. :slightly_smiling_face:

Dobiasd:
Ah, what do you think about `x % 2 != 0` instead of `x & 1`? It too works for negative numbers but requires less knowledge about bit representations.

moowlf:
Sure, I will modify it and add some tests for negative input.

Dobiasd:
Perfect. Thanks a lot again for finding, reporting, and fixing this silly bug. :slightly_smiling_face:

PR #248, opened 2021-09-23
Unify is_odd, is_even, and their usage

Dobiasd:


examples/readme_perf_examples.cpp
```diff
@@ -18,7 +18,7 @@
 #include <string>
 #include <vector>
 
-bool is_odd_int(int x) { return x % 2 == 1; }
+bool is_odd_int(int x) { return x % 2 != 1; }
```
pthom, line 21: Hi, 
May be I'm bike shedding, since only the performance is tested here. But this test for is_odd seems to have an error. Btw, may be `auto is_odd_int = fplus::is_odd<int>` would be a little more succinct.

Dobiasd: Thanks. :+1:
I just fixed it. :slightly_smiling_face:
https://github.com/Dobiasd/FunctionalPlus/commit/b216c089bf92803129cb874b6ab4e23690244405

I agree, `auto is_odd_int = fplus::is_odd<int>` would be more succinct. But in that case, since it's a performance test, I feel it might be good to directly make visible what is used without an indirection to an "external" definition.

PR #261, opened 2022-05-07
add foreach, foreach_parallel and foreach_parallel_n_threads (fix #260)

### LINKED ISSUE #260: parallel_for ? (no, this is not the 1st of April ;-)
Yes, this is controversial; since it goes against any kind of functional paradigm ;-)

However, sometimes I would like to run a "for loop" in parallel.

Take this for example:

````cpp
int main()
{
    // Motivational example, let's say we have a matrix whose rows we want to fill in parallel

    int rows = 1000, cols = 1500;
    // cv::Mat mandelbrot_image = ...

    auto fill_mandelbrot_row = [/* & mandelbrot_image */](size_t row_idx)
    {
        // ... Fill the elements in mandelbrot_image on this row
    };


    std::vector<int> numbers_squared;


    fplus::parallel_for(fill_mandelbrot_row, 0, rows);
}
````

In that case, we could use this draft:

````cpp
namespace fplus
{

    // F is of type Int -> Io()
    template<typename F, typename Number>
    void parallel_for(F f, Number lo, Number hi) 
    {
        auto range = fplus::numbers(lo, hi);

        auto f_dummy_return = [&f](Number v) 
        {
            f(v);
            return true;
        };

        fplus::transform_parallelly(f_dummy_return, range);
    }


    // F is of type Int -> Io()
    template<typename F, typename Number>
    void parallel_for_n_threads(int nb_threads, F f, Number lo, Number hi) 
    {
        auto range = fplus::numbers(lo, hi);

        auto f_dummy_return = [&f](Number v) 
        {
            f(v);
            return true;
        };

        fplus::transform_parallelly_n_threads(nb_threads, f_dummy_return, range);
    }


}

````

[Demo on compiler explorer](https://godbolt.org/z/8coEjfWKE)

### PR DESCRIPTION
**Pull request by pthom**
See https://github.com/Dobiasd/FunctionalPlus/issues/260

Example usage:

````cpp
 constexpr int rows = 3000, cols = 2000;
 MandelbrotMat<rows, cols> mandelbrot_mat;
 auto fill_mandelbrot_row = [& mandelbrot_mat ](int row_idx) {
        ...
 };
 mandelbrot_mat = MandelbrotMat<rows, cols>();
 auto rows_idxs = fplus::numbers(0, rows);
 fplus::parallel_for(fill_mandelbrot_row, rows_idxs);
````

### DIFF
--- a/include/fplus/side_effects.hpp
+++ b/include/fplus/side_effects.hpp
@@ -283,6 +283,48 @@ auto execute_n_times(std::size_t n, const Effect& eff)
     }
 }
 
+// API search type: for_each : (Io a, [a]) -> Io ()
+// Runs the function `f` on all the container elements.
+// The function will perform its side effects, and nothing is returned.
+template<typename F, typename Container>
+void for_each(F f, const Container& xs)
+{
+    using IdxType = typename Container::value_type;
+    auto f_dummy_return = [&f](const IdxType& v) {
+        f(v);
+        return true;
+    };
+    fplus::transform(f_dummy_return, xs);
+}
+
+// API search type: parallel_for_each : (Io a, [a]) -> Io ()
+// Runs the function `f` in parallel on all the container elements.
+// The function will perform its side effects, and nothing is returned.
+template<typename F, typename Container>
+void parallel_for_each(F f, const Container& xs)
+{
+    using IdxType = typename Container::value_type;
+    auto f_dummy_return = [&f](const IdxType& v) {
+        f(v);
+        return true;
+    };
+    fplus::transform_parallelly(f_dummy_return, xs);
+}
+
+// API search type: parallel_for_each_n_threads : (Int, Io a, [a]) -> Io ()
+// Runs the function `f` in parallel on all the container elements, using `n_threads` threads.
+// The function will perform its side effects, and nothing is returned.
+template<typename F, typename Container>
+void parallel_for_each_n_threads(size_t n_threads, F f, const Container& xs)
+{
+    using IdxType = typename Container::value_type;
+    auto f_dummy_return = [&f](const IdxType& v) {
+        f(v);
+        return true;
+    };
+    fplus::transform_parallelly_n_threads(n_threads, f_dummy_return, xs);
+}
+
 // API search type: execute_serially_until_failure : [Io Bool] -> Io Bool
 // Returns a function that (when called) executes the given side effects
 // one after another until one of them returns false.
--- a/test/side_effects_test.cpp
+++ b/test/side_effects_test.cpp
@@ -76,3 +76,24 @@ TEST_CASE("side_effects_test - execute_parallelly_atomic")
     execute_parallelly(replicate(4, inc_atomic_int_return_true))();
     REQUIRE_EQ(atomic_int.load(), 4);
 }
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: include/fplus/side_effects.hpp [hunk 1]
```diff
@@ -283,6 +283,46 @@ auto execute_n_times(std::size_t n, const Effect& eff)
...
+// constexpr int rows = 3000, cols = 2000;
+// MandelbrotMat<rows, cols> mandelbrot_mat;
+// auto fill_mandelbrot_row = [& mandelbrot_mat ](int row_idx) {
+//        ...
+// };
+// mandelbrot_mat = MandelbrotMat<rows, cols>();
+// auto rows_idxs = fplus::numbers(0, rows);
+// fplus::parallel_for(fill_mandelbrot_row, rows_idxs);
+template<typename F, typename ContainerIdxs>
+void parallel_for(F f, const ContainerIdxs& idxs_to_iterate)
```
Dobiasd, line 301: I wouldn't even call it `ContainerIdxs& idxs_to_iterate`, but just `Container& xs`. Of course, the elements can be indexes, but they don't have to. :)

test/side_effects_test.cpp
```diff
@@ -76,3 +76,49 @@ TEST_CASE("side_effects_test - execute_parallelly_atomic")
...
+{
+    std::vector<int> nb_iterations;
+    MandelbrotMat() : nb_iterations(rows * cols, 0) {}
+};
+
+TEST_CASE("side_effects_test - parallel_for")
+{
+    using namespace fplus;
+
+    constexpr int rows = 3000, cols = 2000;
```
Dobiasd, line 91: If I understand correctly, on a 64-bit system, this will use 48 MB. How about a smaller, maybe even generally simpler, test?

include/fplus/side_effects.hpp
```diff
@@ -283,6 +283,46 @@ auto execute_n_times(std::size_t n, const Effect& eff)
...
+    };
+    fplus::transform_parallelly(f_dummy_return, idxs_to_iterate);
+}
+
+// API search type: parallel_for : (Io a, [a]) -> Io ()
+// Runs the function `f` in parallel on all the `idxs_to_iterate` container elements,
+// using `n_threads` threads.
+// The function will perform its side effects, and nothing is returned.
+template<typename F, typename ContainerIdxs>
+void parallel_for_n_threads(size_t n_threads, F f, const ContainerIdxs& idxs_to_iterate)
```
Dobiasd, line 316: What do you think about naming there functions
- `parallel_for_each`
- `parallel_for_each_n_threads`

instead?

It expresses the "execute for each element in this generic container", and also seems to be kind of a default name for such a function.

We could even introduce a normal (non-parallel) `fplus::for_each` too, so that it would align with the existing three `fplus::transform ...` functions. :slightly_smiling_face:

pthom:
Yes to all your comments !

So, I changed the names to for_each, for_each_parallel and for_each_parallel_n_threads.

I removed the example from the API doc, as I 'm not sure it was that interesting.

Dobiasd:
Thanks, looks good to me now. :+1:

(Since the example is now in the commit message, it won't be lost completely. ^^)

The original test was also interesting (at least to me), but the new one feels more standard, simpler, and boring. Boring code is good code. :grin:

PR #262, opened 2022-05-08
foreach: update include_all_in_one

pthom:
Oops, I had forgotten to commit the all in one header, sorry!

Also, I added two entries to .gitignore (those are CLion build artifacts, the IDE I use most of the time).
Is that OK?

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ api_search/frontend/elm-stuff
 api_search/frontend/deploy.sh
 api_search/frontend/src/Database.elm
 .mypy_cache
+.idea
+cmake-build-*/
--- a/include_all_in_one/include/fplus/fplus.hpp
+++ b/include_all_in_one/include/fplus/fplus.hpp
@@ -13662,6 +13662,48 @@ auto execute_n_times(std::size_t n, const Effect& eff)
     }
 }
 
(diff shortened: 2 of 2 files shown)


Dobiasd:
Good catch. I had also forgotten about this header. :woozy_face:

I've now also updated the API search: https://i.imgur.com/oIxx2B8.png

And having these things in the `.gitignore` makes sense, I think. :+1:

Dobiasd:
Oh, it seems, no lines in `fwd_instances.autogenerated_defines` are generated. Do you have an idea why?

pthom:
Oh, good catch!
Hum, by looking at `gather_function_names_and_bind_counts` it seems I forgot to add a line `fwd bind count`.

````py

def gather_function_names_and_bind_counts():
    """
    Looks for pair of lines that look like this in the amalgamation content:
    
        API search type: flatten_tree_breadth_first : Tree a -> [a]
        fwd bind count: 0
    
    and returns a list  [(function_name1, bind_count1), (function_name2, bind_count2), ...]    
    """
   ... 
    for line_number, line_content in enumerate(amalgamated_lines):
        if "fwd bind count" in line_content:                                                             
           ...

````

pthom:
I'll post a PR, I had forgotten about this

Issue #269, opened 2022-10-26
Function to print std::tuple


ferdymercury:
Hello, I was wondering if there is already a simple way of printing out the contents of an std::tuple using fplus.

Or if you would consider adding some of these solutions to your lib: https://stackoverflow.com/questions/6245735/pretty-print-stdtuple

Thanks for your library!

Dobiasd:
Hi, thanks for the positive feedback and the excellent idea. :slightly_smiling_face:

So far, there is nothing in FunctionalPlus to print tuples.

To keep it generic (and reuse the separator/prefix/suffix logic from the `fplus::show_cont...` function family), I guess it would make sense not to add something to directly print a tuple but to just convert it into a list of strings.

Would you like to give the implementation a try in a pull request?

Otherwise, I'd see if I can do something, based on the [cppreference example](https://en.cppreference.com/w/cpp/utility/tuple/tuple_cat).

ferdymercury:
Thanks for the swift reply.

> To keep it generic (and reuse the separator/prefix/suffix logic from the `fplus::show_cont...` function family), I guess it would make sense not to add something to directly print a tuple but to just convert it into a list of strings.

You mean using stringstream instead of iostream ?

ferdymercury:
This could do the job:

```cpp
#include <iostream>
#include <tuple>
#include <string>
#include <sstream>
#include <list>

template<class Tuple, std::size_t N>
struct TuplePrinter {
    static void print(const Tuple& t, std::list<std::string>& sl) 
    {
        TuplePrinter<Tuple, N-1>::print(t,sl);
        std::stringstream ss;
        ss << std::get<N-1>(t);
        sl.emplace_back(ss.str());
    }
};
 
template<class Tuple>
struct TuplePrinter<Tuple, 1> {
    static void print(const Tuple& t, std::list<std::string>& sl) 
    {
        std::stringstream ss;
        ss << std::get<0>(t);
        sl.emplace_back(ss.str());
    }
};
 
template<typename... Args, std::enable_if_t<sizeof...(Args) == 0, int> = 0>
void print(const std::tuple<Args...>& t, std::list<std::string>& sl)
{
    return;
}

template<typename... Args, std::enable_if_t<sizeof...(Args) != 0, int> = 0>
std::list<std::string> print(const std::tuple<Args...>& t)
{
    std::list<std::string> sl;
    TuplePrinter<decltype(t), sizeof...(Args)>::print(t,sl);
[...]

PR #270, opened 2022-10-29
add a way to serialize a tuple as list of strings

ferdymercury:
Fixes https://github.com/Dobiasd/FunctionalPlus/issues/269

### DIFF
--- a/include/fplus/show.hpp
+++ b/include/fplus/show.hpp
@@ -318,7 +318,7 @@ std::list<std::string> stream(const std::tuple<Args...>& t)
 {
     std::list<std::string> sl;
     TupleStreamer<decltype(t), sizeof...(Args)>::stream(t,sl);
-    return std::move(sl);
+    return sl;
 }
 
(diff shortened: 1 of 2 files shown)


Dobiasd:
Looks good. :+1: Thanks! :heart:


There are two `unused parameter` errors in the CI:

```
/__w/FunctionalPlus/FunctionalPlus/include/fplus/show.hpp:306:40: error: unused parameter 't' [-Werror,-Wunused-parameter]
void stream(const std::tuple<Args...>& t, std::list<std::string>& sl)
                                       ^
/__w/FunctionalPlus/FunctionalPlus/include/fplus/show.hpp:306:67: error: unused parameter 'sl' [-Werror,-Wunused-parameter]
void stream(const std::tuple<Args...>& t, std::list<std::string>& sl)
                                                                  ^
```

If you fix them (`void foo(T bar)` -> `void foo(T)`) we're good to merge.

PR #279, opened 2023-09-14
Add missing <cstdint> includes

tocic:
Fixes the following compilation errors in clang15 & gcc13:
- error: no type named 'uint64_t' in namespace 'std'
- error: ‘uint8_t’ is not a member of ‘std’

### DIFF
--- a/include/fplus/container_properties.hpp
+++ b/include/fplus/container_properties.hpp
@@ -18,6 +18,7 @@
 #include <fplus/internal/invoke.hpp>
 
 #include <algorithm>
+#include <cstdint>
 #include <numeric>
 #include <type_traits>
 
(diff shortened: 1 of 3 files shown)


Dobiasd:
Oh, thanks a lot! :+1:

And idea why the [Clang15 test in our CI succeeds even without that change](https://github.com/Dobiasd/FunctionalPlus/actions/runs/6175935119/job/16763814109)?

tocic:
> And idea why the [Clang15 test in our CI succeeds even without that change](https://github.com/Dobiasd/FunctionalPlus/actions/runs/6175935119/job/16763814109)?

Sorry, the problem was reproduced with libstdc++ 13.2, regardless of the compiler. Clang 15 per se is fine - the standard library is what matters.

Dobiasd:
Just released a new version. :slightly_smiling_face:
https://github.com/Dobiasd/FunctionalPlus/releases/tag/v0.2.20-p0

PR #283, opened 2023-10-04
update all_in_one / add missing <cstdint>

pthom:
It seems that the all_in_one header was not updated after the last modifications.

Hello Tobias,

It seems that this simple program can fail to compile on the latest compilers (gcc and clang), and a [fix](https://github.com/Dobiasd/FunctionalPlus/pull/279) that adds `#include <cstdint>` is not included in it.

You can see it on Compiler Explorer:  https://godbolt.org/z/fGEh49crd

So, I just re-ran generate/auto_generate.py for this PR.

Cheers,
Pascal


Dobiasd:
Oh, right. I forgot to do this. :grimacing: Thanks a lot! :heart:

Issue #288, opened 2023-12-25
Methods directly on containers (e.g `maybe`)


tom91136:
I was wondering would adding a few more methods to some of the containers would help with productivity.
For example:

```cpp
    auto o = fplus::maybe<int>(1);
    auto p = fwd::apply(
        o,
        fwd::and_then_maybe([](auto x) { return x > 0 ? fplus::just(1) : fplus::nothing<int>(); }),
        fwd::lift_maybe([](auto x) { return x * 2; }),
        fwd::just_with_default(42)
    );

    auto p2 = o
            .and_then([](auto x) { return x > 0 ? fplus::just(1) : fplus::nothing<int>(); })
            .lift([](auto x) { return x * 2; })
            .get_with_default(42);
```
That way:
 * It aligns with the C++23 monadic interfaces on optional
 * We can drop the `_maybe` suffix as it's clear based on the method receiver
 * We should still keep the free functions but those simply delegates to the instance implementation, so we don't need to 
forward declare it or do anything awkward.
 * For most cases (e.g chaining on the same functor), we can get the same effect of using fwd but without the extra boilerplate. 

I'm happy to open a PR if we think this is a good idea.

Dobiasd:
Oh, that looks like a splendid idea to me in terms of code readability. :heart:

Yeah, we should keep the free functions (to not break old client code). And, I agree, letting the free functions forward to the member functions (instead of the other way around) is the way to go, exactly because of the reason you mentioned. :+1:

I'd be happy to receive your PR. :slightly_smiling_face:

PR #289, opened 2023-12-25
Add monadic (fluent) interface for maybe

tom91136:
This PR implements #288.

I've added new checks directly below the existing maybe tests to make sure it compiles.
An extra test on chaining the fluent interface is also included, we can now do:

```cpp
maybe<int> m(42);
m
  .just_if([](auto x) { return x > 0; })
  .and_then([](auto x) { return x % 2 == 0 ? just<std::string>("a") : nothing<std::string>(); })
  .lift([](auto x) { return just(x); })
  .join()
  .lift_def(nothing<std::string>(), [](auto x) { return just(x + "b"); })
  .get_with_default("c"); // == "ab"
```

I've made the following additions to the API:

 * A `just_if` method (e.g `filterM` for our maybe functor) which I believe is missing.
 * Moved `is_maybe` from `internal` to top-level as we need this for join, and I also think this is a very useful API to expose.


Dobiasd:
That looks splendid! Working with `Maybe` types is much more convenient that way. :heart_eyes:

The only potential downside of the member functions over the free functions might be the `auto` return type on some. However, adding an explicit return type could be quite complex with the derivation based on the passed `F`, etc. So let's keep it that way (with `auto`). In case some users don't like it because the auto-completion in their IDE does not play too well with this, they can still use the free functions (which have explicit return types). :white_check_mark:

Dobiasd:
> A `just_if` method (e.g `filterM` for our maybe functor) which I believe is missing.

That's a very nice addition! :rocket:

Dobiasd:
> Moved `is_maybe` from `internal` to top-level as we need this for join, and I also think this is a very useful API to expose.

Since we don't expose such helpers in other parts of the library, I'm not yet sure about exposing `is_maybe` to the `fplus` namespace here. `.join` (`and` .`and_then`) could also use `is_maybe` if `is_maybe` would still be defined in `internal` (at the same location in the file, i.e., above the class definition). :thinking:

tom91136:
Sure, I've reverted that part.

Dobiasd:
Cool. :+1:

And, thanks a lot for this awesome contribution! :heart:

PR #297, opened 2024-04-04
Added noexcept variants to function traits

LuSo58:
When using `function_traits` with callable items that are noexcept, no specialization would match. Here is a simple reproduction case for this.
```cpp
#include <iostream>
#include <functional>

#include "fplus/fplus.hpp"

template<typename F, typename T = typename fplus::utils::function_traits<F>::template arg<0>::type>
auto forward(F f) {
  return [f](auto x) {
    auto ret = f(x);
    return ret;
  };
}

class klass {
  public:
    int x;

    int normal() const {
      return x;
    }

    int noexc() const noexcept {
      return x;
    }
};

int half(int x) {
  return x / 2;
}

int half_noexc(int x) noexcept {
  return x / 2;
}

struct functor {
  int operator()(int x) {
    return x + 5;
  }
};

struct functor_noexc {
  int operator()(int x) noexcept {
    return x + 5;
  }
};

int main() {
  klass k = {42};
  std::cout << forward(std::mem_fn(&klass::normal))(k) << std::endl; // OK
  std::cout << forward(std::mem_fn(&klass::noexc))(k) << std::endl; // Will not compile
  std::cout << forward(half)(42) << std::endl; // OK
  std::cout << forward(half_noexc)(42) << std::endl; // Will not compile
  std::cout << forward(functor())(42) << std::endl; // OK
  std::cout << forward(functor_noexc())(42) << std::endl; // Will not compile
  auto lambda = [](int x) {
    return x + 10;
  };
  auto lambda_noexc = [](int x) noexcept {
    return x + 10;
  };
  std::cout << forward(lambda)(42) << std::endl; // OK
  std::cout << forward(lambda_noexc)(42) << std::endl; // Will not compile
}
```

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -37,7 +37,7 @@ tests requires [doctest](https://github.com/doctest/doctest).
 
 Install doctest:
 ```bash
-git clone --depth=1 --branch=v2.4.9 https://github.com/doctest/doctest
+git clone --depth=1 --branch=v2.4.11 https://github.com/doctest/doctest
 cmake -S doctest -B doctest/build -DDOCTEST_WITH_TESTS=OFF -DDOCTEST_WITH_MAIN_IN_STATIC_LIB=OFF
 cmake --build doctest/build -j 4
 sudo cmake --install doctest/build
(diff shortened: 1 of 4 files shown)


LuSo58:
I added macros to hide the noexcept specs from compilers, that don't seem to distinguish between the signatures. See the checks results. I don't know how to force rerun the checks, so I can't check if it works easily.

Dobiasd:
Thanks!

Yeah, I need to approve such runs - probably to prevent people from using the CI to mine Bitcoins or similar. ;)

Issue #301, opened 2024-05-22
container_common_test.: out of bounds warning on GCC 14


offa:
Compiling fplus on GCC 14 issues an out of bounds warning:

```
 [ 23%] Building CXX object CMakeFiles/function_traits_test.dir/function_traits_test.cpp.o
In file included from /usr/local/include/c++/14.1.0/deque:62,
                 from /__w/FunctionalPlus/FunctionalPlus/test/container_common_test.cpp:7:
In static member function 'static _Up* std::__copy_move<_IsMove, true, std::random_access_iterator_tag>::__copy_m(_Tp*, _Tp*, _Up*) [with _Tp = const int; _Up = int; bool _IsMove = false]',
    inlined from '_OI std::__copy_move_a2(_II, _II, _OI) [with bool _IsMove = false; _II = const int*; _OI = int*]' at /usr/local/include/c++/14.1.0/bits/stl_algobase.h:521:30,
    inlined from '_OI std::__copy_move_a1(_II, _II, _OI) [with bool _IsMove = false; _II = const int*; _OI = int*]' at /usr/local/include/c++/14.1.0/bits/stl_algobase.h:548:42,
    inlined from '_OI std::__copy_move_a(_II, _II, _OI) [with bool _IsMove = false; _II = __gnu_cxx::__normal_iterator<const int*, vector<int> >; _OI = int*]' at /usr/local/include/c++/14.1.0/bits/stl_algobase.h:555:31,
    inlined from '_OI std::copy(_II, _II, _OI) [with _II = __gnu_cxx::__normal_iterator<const int*, vector<int> >; _OI = int*]' at /usr/local/include/c++/14.1.0/bits/stl_algobase.h:651:7,
    inlined from 'static _ForwardIterator std::__uninitialized_copy<true>::__uninit_copy(_InputIterator, _InputIterator, _ForwardIterator) [with _InputIterator = __gnu_cxx::__normal_iterator<const int*, std::vector<int> >; _ForwardIterator = int*]' at /usr/local/include/c++/14.1.0/bits/stl_uninitialized.h:147:27,
    inlined from '_ForwardIterator std::uninitialized_copy(_InputIterator, _InputIterator, _ForwardIterator) [with _InputIterator = __gnu_cxx::__normal_iterator<const int*, vector<int> >; _ForwardIterator = int*]' at /usr/local/include/c++/14.1.0/bits/stl_uninitialized.h:185:15,
[...]

PR #302, opened 2024-05-22
Add GCC 14 CI build

offa:
Draft as there's a warning issued #301 which needs further investigation.


offa:
Thanks @Dobiasd, GCC 14 looks fixed now! :+1:

PR #307, opened 2024-06-22
Switch to the default configuration of CodeQL

Dobiasd:
It's no longer needed to have a custom configuration file:
https://github.blog/changelog/2023-01-09-code-scanning-can-be-set-up-more-easily-without-committing-a-workflow-file-to-the-repository/

So let's try the default configuration. If it turns out it's not good for us, we can switch back.


Dobiasd:
Switched back to the custom config. Without it, CodeQL does not know how to build the code.

PR #318, opened 2025-06-13
changed fplus::internal::get_index to reduce templated recursion and …

PraisePancakes:
…introduced contains metafunction to test whether type list contains a type

### DIFF
--- a/include/fplus/variant.hpp
+++ b/include/fplus/variant.hpp
@@ -93,23 +93,20 @@ namespace internal {
         typedef std::shared_ptr<T> type;
     };
 
-    // http://stackoverflow.com/a/27588263/1866775
-
-    template <typename T, typename... Ts>
-    struct get_index;
-
-    template <typename T, typename... Ts>
-    struct get_index<T, T, Ts...> : std::integral_constant<std::size_t, 0> {
+    template <typename What, typename... Ts>
+    struct contains {
+        /// True if T is in Ts...
+        constexpr static bool value = (std::is_same_v<What, Ts> || ...);
     };
 
-    template <typename T, typename Tail, typename... Ts>
-    struct get_index<T, Tail, Ts...> : std::integral_constant<std::size_t, 1 + get_index<T, Ts...>::value> {
-    };
-
-    template <typename T>
-    struct get_index<T> {
-        // condition is always false, but should be dependant of T
-        static_assert(sizeof(T) == 0, "element not found");
+    template <typename What, typename... Ts>
+    struct get_index {
+        /// Index of type T in List.
+        constexpr static auto index = []() {
+            return (contains<What, Ts...>::value ? []() {signed i = 0;
+            (... && (!std::is_same_v<What, Ts...> && ++i));
+            return i; }() : -1);
+        }();
     };
 
     template <typename T, typename... Ts>
--- a/include_all_in_one/include/fplus/fplus.hpp
+++ b/include_all_in_one/include/fplus/fplus.hpp
@@ -3295,7 +3295,7 @@ namespace internal {
         std::size_t pos_;
     };
 
-#if defined(_MSC_VER) && _MSC_VER >= 1900
+#if defined(_MSC_VER) && _MSC_VER >= 1900 && _MSC_VER < 1915
     template <typename T, std::size_t N>
     struct std::_Is_checked_helper<array_back_insert_iterator<T, N>>
         : public true_type { // mark array_back_insert_iterator as checked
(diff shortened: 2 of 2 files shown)


Dobiasd:
Thanks, that looks nicely concise! :heart:

But is it C++14-compatible? (Asking because the CI is failing.)

PraisePancakes:
> Thanks, that looks nicely concise! :heart:
> 
> 
> 
> But is it C++14-compatible? (Asking because the CI is failing.)

Ahh unfortunately I dont know I only tested on 20 & 23 :( ill test on 14 when I get home and get back to you!

Dobiasd:
Ah, ok. I mean, we can increase the compiler requirements for FunctionalPlus if we have a strong reason. Only quite a few users should be stuck with C++14.

What problem did the change you implemented solve for you?

Dobiasd:
Yes, I do too. :) I just wanted to make sure I fully understand the intention of the proposed change.

Did you perform any benchmarking on compile times comparing the old version vs. the new one?

PraisePancakes:
Yes I did! I tested on a list of roughly 400 types, and the fold expression was faster by roughy 3-4 milliseconds. Though this may be a small difference I tested it in a brand new environment. Feel free to try it yourself and let me know if you see any differing results!

Dobiasd:
Thanks, but it still [does neither work in the CI](https://github.com/Dobiasd/FunctionalPlus/actions/runs/15652941497/job/44118092238?pr=318) nor for me locally.

Here is a minimal example to reproduce the problem (it's an excerpt of the automated test in this repo): https://gist.github.com/Dobiasd/90bea43eb723d87c18cc9e5cc9b1c290
Could you please test with this one? (It does work with the latest version of the `master` branch, but with your updated `variant.hpp`, I get the same errors that we see in the PR CI.)

PraisePancakes:
Hello, sorry for the late reply, I just got home from work not too long ago. I managed to get the CI to comply with every complier except for MSVC, but the reason seems to be due to function traits. Can you spot a reason as to why that is? Also, the previous commit wasn't cooperating because of a naming error with the get_index definition,  (switched it from "index" to "value").

Dobiasd:
C++14 should work fine for `test/function_traits_test.cpp` and not produce errors in `function_traits.hpp`.

Maybe MSCV is falling back to something even older now because it does not understand `cxx_std_23`.

According to [this page](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170), it [does not yet](https://i.imgur.com/5vwPoZA.png) support `/std:c++23`, only `/std:c++20` and `/std:c++23preview`.

Would your changes also work with `cxx_std_20` instead of `cxx_std_23`?

PraisePancakes:
Yes the changes should absolutely work with c++20, thats the version i tested in with a local environment and fold expressions were introduced in c++17!

Dobiasd:
Thanks, ok, I see.

We won't drop MSVC support of FunctionalPlus for this, so it would be cool if you could come up with a way to make it work with this compiler too.

And it would be cool if it does not require C++23. C++20 might be ok, but C++17 would be better. Ideally, I'd like to lift the requirement up from C++14 as little as possible, because FunctionalPlus is used in frugally-deep, and frugally-deep is used by some projects that compile for weird (sometimes old) architectures, so they might not have the option to use the latest compilers.

PraisePancakes:
> Thanks, ok, I see.
> 
> 
> 
> We won't drop MSVC support of FunctionalPlus for this, so it would be cool if you could come up with a way to make it work with this compiler too.
> 
> 
> 
> And it would be cool if it does not require C++23. C++20 might be ok, but C++17 would be better. Ideally, I'd like to lift the requirement up from C++14 as little as possible, because FunctionalPlus is used in frugally-deep, and frugally-deep is used by some projects that compile for weird (sometimes old) architectures, so they might not have the option to use the latest compilers.

Thats totally fine, ill keep working on it and ill keep in touch with you if I can find the issue, thanks for your communication! Keep up the good work!

Dobiasd:
@PraisePancakes Any update on that?

PraisePancakes:
> @PraisePancakes Any update on that?

Unfortunately I could not find the problem to resolve the issue with msvc, I also had university come up and research for our Comp. Sci. department so I did not resume in the search for the issue. I did however get inspired by your library and developed a more modern approach to a functional library.

Dobiasd:
> Unfortunately I could not find the problem to resolve the issue with msvc, I also had university come up and research for our Comp. Sci. department so I did not resume in the search for the issue.

Ok, no problem. I'll close this PR for now, then. :+1:

> I did however get inspired by your library and developed a more modern approach to a functional library.

That's awesome! :rocket:

I just had a look at your [endo](https://github.com/PraisePancakes/endo), and it looks great!
- modern recursive lambda approach for currying
- cleaner storage in the "maybe" type using `std::variant`
- `std::remove_cvref_t` (C++20) for function traits
- simple(r) recursive variadic templates for composition

And you even have a Y-combinator. :grin:

PR #326, opened 2026-02-26
Add Clang 22 CI build

offa:


offa:
doctest causes an error under clang‑22 (https://github.com/doctest/doctest/issues/1042).

Dobiasd:
Thanks a lot for reporting in the `doctest` repo! :heart:

Since we have no urgency here, let's see when they will fix it, so we don't need a workaround like this in our `cmake/warnings.cmake`:

```
    # See https://github.com/doctest/doctest/issues/1042
    if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
        CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 22)
        list(APPEND project_warnings -Wno-c2y-extensions)
    endif ()
```

:see_no_evil:

offa:
This is the only warning, so users can disable the warnings on their side without any other problems I guess.

Dobiasd:
Users of FunctionalPlus probably won't even need to disable the warning, because doctest is only imported in the test files, not in the actual library headers.

offa:
> Users of FunctionalPlus probably won't even need to disable the warning, because doctest is only imported in the test files, not in the actual library headers.

Ah yes, that makes sense. So even less urgent.

Dobiasd:
But I guess we need to disable it here to make the CI of this PR pass. Or what approach would you suggest?

offa:
I’m not in a hurry with the PR, and it will still take some time before Clang 22 is widely adopted.

Dobiasd:
Ah, right. Yeah, then let's just wait. :relaxed:

offa:
Quick update: Doctest fixed the problem and we now just have to wait for a new release.

offa:
New release is out – with Clang 22 support :tada:.

Conan package of doctest 2.5.0 is not yet available, but ticket open: https://github.com/conan-io/conan-center-index/issues/29875

However, I still need to fix a macOS issue to get back to a green build:

```
test/udemy_course_test.cpp:315:23: error: reference to 'user' is ambiguous
  315 |     const std::vector<user> users = {
      |                       ^
```

Mac OS picks up a `user` type from some internal XCode SDK headers :man_shrugging:.

test/udemy_course_test.cpp
```diff
@@ -291,18 +291,18 @@ TEST_CASE("udemy_course_test - Function_composition")
 }
 
 namespace Programming_challenge_an_SQL_analogy {
-struct user {
+struct User {
```
offa, line 294: A not particularly complex approach 😅, but it avoids the symbol ambiguity with internal Xcode SDK headers.

Dobiasd: Lol 😁👍

offa:
Final review and we are good to go I hope :-)

Dobiasd (approved):

Dobiasd:
Perfect, thanks a lot! ❤️

PR #327, opened 2026-02-28
Scripts: Fail on error

offa:
Fail the scripts on error to avoid silently ignoring.

### DIFF
--- a/script/ci.sh
+++ b/script/ci.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 # Central script called by the CI
 # Usage: 
 #    ci.sh {run_build|run_tests}
--- a/script/ci.sh
+++ b/script/ci.sh
@@ -19,7 +19,7 @@ if [[ "$OSTYPE" == "msys" || "$OSTYPE" == "cygwin" ]]; then
     # Windows path handling is a never ending source of ...
     INSTALL_PREFIX="$USERPROFILE\\.local"
     INSTALL_PREFIX=$(cygpath -u "$INSTALL_PREFIX")  # Make CMake happy when using git bash under windows
-    export CMAKE_PREFIX_PATH=$INSTALL_PREFIX;$CMAKE_PREFIX_PATH  # Notice the ";" instead of ":"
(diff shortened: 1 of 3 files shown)


script/ci.sh
```diff
@@ -26,7 +28,7 @@ fi
 # Function to install doctest
 _install_doctest() {
     cd "$REPO_DIR"
-    git clone --depth=1 --branch=v2.4.11 https://github.com/doctest/doctest
+    git clone --depth=1 --branch=v2.4.12 https://github.com/doctest/doctest
```
offa, line 31: Fixes this Mac OS error present in current builds:

```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
-- Configuring incomplete, errors occurred!
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


Error: could not find CMAKE_PROJECT_NAME in Cache
```

script/ci.sh
```diff
@@ -17,7 +19,7 @@ if [[ "$OSTYPE" == "msys" || "$OSTYPE" == "cygwin" ]]; then
     # Windows path handling is a never ending source of ...
     INSTALL_PREFIX="$USERPROFILE\\.local"
     INSTALL_PREFIX=$(cygpath -u "$INSTALL_PREFIX")  # Make CMake happy when using git bash under windows
-    export CMAKE_PREFIX_PATH=$INSTALL_PREFIX;$CMAKE_PREFIX_PATH  # Notice the ";" instead of ":"
+    export CMAKE_PREFIX_PATH="${INSTALL_PREFIX}${CMAKE_PREFIX_PATH:+;${CMAKE_PREFIX_PATH}}" # Notice the ";" instead of ":"
```
offa, line 22: Related error in Windows builds:

```
script/ci.sh: line 20: /c/Users/runneradmin/.local: No such file or directory
```

Dobiasd:
Ah, very good findings. :+1: Thanks a lot. :heart:

